PR #6, opened 2022-07-18
Fix Azure template

mpagot:
Restore the comments about general variables
Restore admin_user as cloudadmin
Add a note about the ssh private key


mr-stringer (commented):
Looks good!

PR #7, opened 2022-07-18
Bring libvirt provider to working state

lpalovsky:
This PR removes salt provisioner from libvirt. 
Couple of important parameters are added into tfvars example file.
Faulty network output preventing deployment finish is disabled until fixed.


terraform/libvirt/outputs.tf
```diff
@@ -4,9 +4,9 @@ output "cluster_nodes_ip" {
   value = module.hana_node.output_data.private_addresses
 }
 
-output "cluster_nodes_public_ip" {
-  value = module.hana_node.output_data.addresses
-}
+#output "cluster_nodes_public_ip" {
```
mpagot, line 7: Is it not one of the important values needed to be exported in the Ansible inventory? Is it not relevant for libvrt?

lpalovsky: Yes it is but currently it is broken and blocking deployment to finish. Since libvirt is not a priority I left it out till it is fixed.

mr-stringer (approved):
Looks good to me

PR #8, opened 2022-07-18
GitHub actions for terraform files

mpagot:
It runs `terraform fmt` and `terraform validate` once on each provider folder (for the moment only azure).


mr-stringer (approved):
Excellent.  Looks good.

PR #31, opened 2022-07-28
Add arguments to build.sh and destroy.sh

mpagot:
build and destroy glue sripts support new arguments:
-k to provide th essh key path
-q to tune the output for openQA
-s to skip the Ansible part
Optimize the Dockerfile by removing a layer and to pipe curl to tar
Glue script to POSIX (an dno more bash)
Change the readme to document the new arguments
Add awscli to the requirements.txt file


Dockerfile
```diff
@@ -1,11 +1,11 @@
 # syntax=docker/dockerfile:experimental
 
 FROM opensuse/tumbleweed:latest
```
mr-stringer, line 3: Should we use latest or stick to a known release?  If something changes and breaks our automation, we'd most likely blame the automation.

Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-RUN zypper ref && zypper up -y
```
mr-stringer, line 4: Same here, it won't give us a stable environment.

Dockerfile
alvarocarvajald: Known release has my vote.

alvarocarvajald:
I am not sure whether `terraform/azure/terraform.tfvars.openqa` should be in this repo or in os-autoinst-distri-opensuse. IMO it makes more sense to centralize openQA related assets in the openQA repositories.

mpagot:
> I am not sure whether `terraform/azure/terraform.tfvars.openqa` should be in this repo or in os-autoinst-distri-opensuse. IMO it makes more sense to centralize openQA related assets in the openQA repositories.

Content of this file (list of variables like **%DEPLOYMENTNAME%** that openQA will replace with proper values) has to be known by openQA test code (or library). 

On the other hand, the content of this .tfvars.openqa has to be aligned to any new variables introduced by the qe-sap-deployment, as anyone is touching the .tfvars.example is supposed to touch this one too (that otherwise is sort of hidden in another repo)

alvarocarvajald:
> On the other hand, the content of this .tfvars.openqa has to be aligned to any new variables introduced by the qe-sap-deployment, as anyone is touching the .tfvars.example is supposed to touch this one too (that otherwise is sort of hidden in another repo)

Yes, that's a valid counterpoint for having it here.

PR #24, opened 2022-07-25
Remove salt provisioner and use of IAM roles

alvarocarvajald:
This PR removes the terraform modules which reference the salt provisioning and the use of IAM roles from the base configuration imported from https://github.com/SUSE/ha-sap-terraform-deployments.

Tested with:
```
cd qe-sap-deployment/terraform/aws
cp terraform.tfvars.example terraform.tfvars
patch terraform.tfvars ~/tfvars.patch
terraform init
terraform workspace new myworkspace
terraform plan
terraform apply --auto-approve
ssh -i ~/.ssh/id_rsa ec2-user@<X.X.X.X>
terraform destroy --auto-approve
```
Tests were done by applying the following patches to the existing `terraform/aws/terraform.tfvars.example`:

1. Changes to deploy 2 HANA nodes:
```
--- qe-sap-deployment/terraform/aws/terraform.tfvars.example    2022-07-28 11:18:20.604833790 -0400
+++ terraform.tfvars    2022-07-28 11:19:45.288833790 -0400
@@ -38,7 +38,7 @@
 #aws_access_key_id = my-access-key-id
 #aws_secret_access_key = my-secret-access-key
 # aws-cli credentials file. Located on ~/.aws/credentials on Linux, MacOS or Unix or at C:\Users\USERNAME\.aws\credentials on Windows
-#aws_credentials = "~/.aws/credentials"
+aws_credentials = "~/.aws/credentials"
 
 # Supply SCC registration code to register BYOS images to SCC. Keep it commented if using PAYG images or if registration of BYOS
 # images is not required
@@ -54,16 +54,16 @@
 
 # Default os_image and os_owner. These values are not used if the specific values are set (e.g.: hana_os_image)
 # BYOS example with sles4sap 15 sp3 (this value is a pattern, it will select the latest version that matches this name)
-#os_image = "suse-sles-sap-15-sp3-byos"
-#os_owner = "amazon"
+os_image = "suse-sles-sap-15-sp3-byos"
+os_owner = "amazon"
 
 # The project requires a pair of SSH keys (public and private) to provision the machines
 # The private key is only used to create the SSH connection, it is not uploaded to the machines
 # Besides the provisioning, the SSH connection for this keys will be authorized in the created machines
 # These keys are provided using the next two variables in 2 different ways
 # Path to already existing keys
-public_key  = "/home/myuser/.ssh/id_rsa.pub"
-private_key = "/home/myuser/.ssh/id_rsa"
[...]

PR #34, opened 2022-08-01
Configure cloudadmin as the admin user for EC2 instances

alvarocarvajald:
This project is already using the `cloudadmin` user as the admin user for remote systems in Azure and GCP, so AWS/EC2 was the outlier by having the `ec2-user` provided by default in EC2 as the admin user. This commit adds `cloudadmin` as the admin user instead of `ec2-user` via the `user_data` option in the AWS EC2 instances configuration. At the same time, `ec2-user` is being dropped from the remote systems as a consequence.

`user_data` configuration for `cloudadmin` as a substitute to `ec2-user`
was taken from: https://aws.amazon.com/premiumsupport/knowledge-center/ec2-user-account-cloud-init-user-data/

Tested with:
```
# cd qe-sap-deployment/terraform/aws
# cp terraform.tfvars.example terraform.tfvars
# patch terraform.tfvars ~/tfvars.patch
# terraform init
# terraform workspace new myworkspace
# terraform plan
# terraform apply --auto-approve
# terraform output -json | jq -rc '.cluster_nodes_public_ip.value, .iscsisrv_public_ip.value'
["X.X.X.X","Y.Y.Y.Y"]
Z.Z.Z.Z
# ssh -i ~/.ssh/id_rsa cloudadmin@X.X.X.X
cloudadmin@ip-10-0-1-10:~> id
uid=1000(cloudadmin) gid=100(users) groups=100(users),1000(wheel)
cloudadmin@ip-10-0-1-10:~> sudo su -
ip-10-0-1-10:~ # exit
# terraform destroy --auto-approve
```
The terraform.tfvars files used were the same as described in #24

### DIFF
--- a/terraform/aws/modules/drbd_node/main.tf
+++ b/terraform/aws/modules/drbd_node/main.tf
@@ -46,6 +46,7 @@ resource "aws_instance" "drbd" {
   vpc_security_group_ids      = [var.security_group_id]
   availability_zone           = element(var.availability_zones, count.index)
   source_dest_check           = false
+  user_data                   = templatefile("${path.root}/cloudadmin.tpl", { publickey = var.common_variables["public_key"] })
 
   root_block_device {
     volume_type = "gp2"
@@ -73,7 +74,7 @@ module "drbd_on_destroy" {
   source       = "../../../generic_modules/on_destroy"
   node_count   = var.drbd_count
   instance_ids = aws_instance.drbd.*.id
-  user         = "ec2-user"
+  user         = "cloudadmin"
   private_key  = var.common_variables["private_key"]
   public_ips   = aws_instance.drbd.*.public_ip
   dependencies = var.on_destroy_dependencies
--- a/terraform/aws/cloudadmin.tpl
+++ b/terraform/aws/cloudadmin.tpl
@@ -1,0 +1,11 @@
+#cloud-config to deploy cloudadmin user in EC2 instances
+cloud_final_modules:
+- [users-groups,always]
+users:
+  - name: cloudadmin
+    groups: [ wheel ]
+    sudo:
+      - "ALL=(ALL) NOPASSWD:ALL"
+    shell: /bin/bash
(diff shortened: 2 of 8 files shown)


alvarocarvajald:
As stated this PR is removing `ec2-user` from the configuration. If we think we will need it, it's a matter of adding it in `cloudadmin.tpl`.

lpalovsky (approved):
Looks good from my side.

terraform/aws/modules/drbd_node/main.tf
```diff
@@ -46,6 +46,7 @@ resource "aws_instance" "drbd" {
   vpc_security_group_ids      = [var.security_group_id]
   availability_zone           = element(var.availability_zones, count.index)
   source_dest_check           = false
+  user_data                   = templatefile("${path.root}/cloudadmin.tpl", { publickey = var.common_variables["public_key"] })
```
mpagot, line 49: Why not to have the username itself as a variable in the template? (maybe as we do not need it...)

```
  user_data                   = templatefile("${path.root}/cloudadmin.tpl", 
{
 username = var.authorized_user,
 publickey = var.common_variables["public_key"]
 })
  ```
  
  And the template like
  ```
  #cloud-config to deploy cloudadmin user in EC2 instances
cloud_final_modules:
- [users-groups,always]
users:
  - name: ${username}
    groups: [ wheel ]
    sudo:
      - "ALL=(ALL) NOPASSWD:ALL"
    shell: /bin/bash
    ssh-authorized-keys: 
    - ${publickey}
  ```

mpagot (approved):
LGTM

terraform/aws/modules/drbd_node/main.tf
alvarocarvajald: Good idea. Didn't think of that as the goal was to add `cloudadmin`, but if it's a variable it can be overwritten via tfvars file or terraform command line. Will do that in a new PR.

PR #36, opened 2022-08-02
Add Ansible inventory generation for all the cloud providers

mpagot:
Tested on Azure and GCP, with both Native and SBD. Test is to call 'terraform apply' and visually inspect the generated inventory.yaml


alvarocarvajald (approved):
LGTM.

Are those files missing the newlines in the last line going to give us problems eventually, or it doesn't matter?

mpagot:
> Are those files missing the newlines in the last line going to give us problems eventually, or it doesn't matter?
Azure and GCP terraform files are checked with `terraform fmt` as part of the PR validation, https://github.com/SUSE/qe-sap-deployment/runs/7635779967?check_suite_focus=true  https://github.com/SUSE/qe-sap-deployment/runs/7635780113?check_suite_focus=true 

the .tmlp will be rendered in a YAML file, that is a Ansible Inventory. I can, at least, confirm that Ansible is able to use them.

PR #40, opened 2022-08-04
ansible linting and VIP corrections

mr-stringer:


mpagot (approved):
LGTM (but I like also to have a GitHub action for that)

mr-stringer:
> LGTM (but I like also to have a GitHub action for that)

https://marketplace.visualstudio.com/items?itemName=redhat.ansible

We're not getting 100% at the moment.  It will take a little refactoring.  But agree that it will be good to have it as a GH action too.

PR #43, opened 2022-08-10
added support for sles-15-sp1

mr-stringer:
* Added support for SLES 15-sp1.
* Cleaned up and standardised the how the cluster state is checked


mpagot (approved):
LGTM

PR #46, opened 2022-08-11
added saptune support

mr-stringer:
saptune support is now added and is the default tuning tool.  sapconf can still be used by setting use_sapconf to 'yes' with sap-hana-preconfigure.yaml.

### DIFF
--- a/ansible/playbooks/sap-hana-preconfigure.yaml
+++ b/ansible/playbooks/sap-hana-preconfigure.yaml
@@ -4,8 +4,11 @@
   become: true
   become_user : root
   vars:
-    scale_out: False
-    use_sapconf: False
+    scale_out: false
+    use_sapconf: false
+    saptune_solution: HANA
+    platform: azure
+    cluster_node: true
     hana_download_path: /hana/shared/install
     url_timeout: 30
     url_retries_cnt: 5
@@ -35,6 +38,9 @@
       when: use_sapconf | bool
 
     # saptune to be handled here with more included tasks!
+    - name: Configure saptune based systems
+      ansible.builtin.include_tasks: ./tasks/saptune.yaml
+      when: not use_sapconf | bool 
     
   handlers:
     - name: reboot
--- a/ansible/playbooks/tasks/saptune.yaml
+++ b/ansible/playbooks/tasks/saptune.yaml
@@ -1,0 +1,79 @@
+---
+- name: Ensure that saptune 3.0 or greater is installed
+  community.general.zypper:
+    name: 'saptune>=3.0'
+    state: present
+
+- name: Ensure conflicting services are stopped and disabled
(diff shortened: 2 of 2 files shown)


ansible/playbooks/sap-hana-preconfigure.yaml
```diff
@@ -4,8 +4,11 @@
   become: true
   become_user : root
   vars:
-    scale_out: False
-    use_sapconf: False
+    scale_out: false
+    use_sapconf: false
+    saptune_solution: HANA
+    platform: azure
```
alvarocarvajald, line 10: Will this be configurable later?

ansible/playbooks/tasks/saptune.yaml
```diff
@@ -0,0 +1,79 @@
...
+    backup: yes
+    owner: root
+    group: root
+    mode: '0640'
+    marker: "<!-- {mark} ANSIBLE MANAGED BLOCK -->"
+    block: |
+     [sysctl]
+     net.ipv4.tcp_timestamps = 0
+     net.ipv4.tcp_tw_reuse = 0
+     net.ipv4.tcp_timestamps = 0
```
alvarocarvajald, line 56: Is it expected that `net.ipv4.tcp_timestamps` is twice in the block? Or is the second one supposed to be `net.ipv4.tcp_tw_recycle`?

ansible/playbooks/tasks/saptune.yaml
```diff
@@ -0,0 +1,79 @@
...
+  when: 
+    - cluster_node
+    - platform == 'azure'
```
alvarocarvajald, line 59: Same question regarding `azure` as above.

ansible/playbooks/sap-hana-preconfigure.yaml
mr-stringer: These are all variables, so yes.  They can be overwritten by calling ansible with the -e flag.  For example, to use sapconf instead of saptune you could do `ansible-playbook -e 'use_sapcong=true /path/to/playbook.yaml`

ansible/playbooks/tasks/saptune.yaml
mr-stringer: It is, yes.  Will fix.

ansible/playbooks/tasks/saptune.yaml
mr-stringer: Yes, variables can be overwritten at run-time.

ansible/playbooks/tasks/saptune.yaml
mr-stringer: Fixed!

alvarocarvajald (approved):
LGTM

PR #50, opened 2022-08-24
Generate tfvars from yaml

lpalovsky:
This PR adds function to generate tfvars file only from yaml config 
values without the need of tfvars template file. It also includes 
pytests with fixes.


mpagot (changes requested):
Just one debug code line to be removed, everything else is fine

mpagot (approved):
LGTM

PR #54, opened 2022-09-01
Fix terrafourm output variables

lpalovsky:
This PR:
- removes some remaining salt grains terraform outputs.
- renames variables from "*cluster*" to "*hana*" to match openqa naming

VR:
* Both test failures are not related to the changes. Terraform deployment finishes in both cases.

Google: https://mordor.suse.cz/tests/4309
Azure: https://mordor.suse.cz/tests/4310


ricardobranco777 (approved):
LGTM

alvarocarvajald (approved):
LGTM

terraform/azure/outputs.tf
```diff
@@ -109,8 +109,8 @@ output "bastion_public_ip" {
 resource "local_file" "ansible_inventory" {
   content = templatefile("inventory.tmpl",
     {
-      hana-name           = module.hana_node.cluster_nodes_name[0],
-      hana-pip            = module.hana_node.cluster_nodes_public_ip[0],
+      hana-name           = module.hana_node.hana_name[0],
```
mpagot, line 112: So we are going to use internally the names already used in the Ansible API (the inventory.yaml) ... right?

mpagot (commented):
LGTM

terraform/azure/outputs.tf
lpalovsky: Yeah iirc.

PR #56, opened 2022-09-01
More Pylint fixes

ricardobranco777:
More Pylint fixes

### DIFF
--- a/scripts/qesap/lib/config.py
+++ b/scripts/qesap/lib/config.py
@@ -97,21 +97,20 @@ def template_to_tfvars(tfvars_template, configure_data):
             return tfvar_content
 
         log.debug("Config has terraform variables")
-        for k, v in configure_data['terraform']['variables'].items():
+        for key, value in configure_data['terraform']['variables'].items():
             key_replace = False
-            # Look for k in the template file content
+            # Look for key in the template file content
             for index, line in enumerate(tfvar_content):
-                match = re.search(k + r'\s?=.*', line)
-                if match:
-                    log.debug("Replace template %s with [%s = %s]", line, k, v)
-                    tfvar_content[index] = f"{k} = {v}\n"
+                if re.search(rf'{key}\s?=.*', line):
+                    log.debug("Replace template %s with [%s = %s]", line, key, value)
+                    tfvar_content[index] = f"{key} = {value}\n"
                     key_replace = True
             # add the new key/value pair
             if not key_replace:
-                log.debug("[k:%s = v:%s] is not in the template, append it", k, v)
-                entry = yaml_to_tfvars_entry(k, v)
+                log.debug("[k:%s = v:%s] is not in the template, append it", key, value)
+                entry = yaml_to_tfvars_entry(key, value)
                 if entry is None:
                     return None
-                tfvar_content.append(entry + '\n')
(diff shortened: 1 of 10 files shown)


scripts/qesap/qesap.py
```diff
@@ -5,6 +5,7 @@
 import os
 import argparse
 import logging
+import select
```
mpagot, line 8: I know about that warning but due to the content, the purpose and complicate nature of the code that need it, I like to keep the import there, hoping in the future to drop them all

scripts/qesap/test/unit/test_qesap_ansible.py
```diff
@@ -370,4 +370,4 @@ def test_ansible_env_config():
 fi
 export ANSIBLE_PIPELINING=True
     """
-    pass
+    return
```
mpagot, line 373: I know about this warning, but it was not so bad remembering me that one more test has to be implemented asap :-)

scripts/qesap/lib/config.py
```diff
@@ -88,7 +88,7 @@ def template_to_tfvars(tfvars_template, configure_data):
         bool: True(pass)/False(failure)
     """
     log.info("Read %s", tfvars_template)
-    with open(tfvars_template, 'r', encoding='utf-8') as filehandler:
+    with open(tfvars_template, encoding='utf-8') as filehandler:
```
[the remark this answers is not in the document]
ricardobranco777: > Why this change? I like the explicit 'r' and 'w' when I go through the code

In this case is not Pylint complaining so it's just a matter of personal taste.  I can drop this commit.

scripts/qesap/qesap.py
ricardobranco777: > I know about that warning but due to the content, the purpose and complicate nature of the code that need it, I like to keep the import there, hoping in the future to drop them all

In this case, Pylint complaints with unused imports.  If we want to keep a near-perfect score with Pylint we should drop these.  We can add them when we really need them after all...

scripts/qesap/test/unit/test_qesap_ansible.py
ricardobranco777: > I know about this warning, but it was not so bad remembering me that one more test has to be implemented asap :-)

I believe it's better to have a TODO comment in this case rather than `pass` which is preferred in other contexts such as try/except.

scripts/qesap/qesap.py
mpagot: I propose to use local pyling suppression 
```
import select # pylint: disable=abstract-class-instantiated
```

scripts/qesap/qesap.py
```diff
@@ -23,20 +24,6 @@
 log = logging.getLogger('QESAPDEP')
 
 
-def os_path_exists(path):
```
mpagot, line 26: This mock was at purpose as indicated in the comment

mpagot:
LGTM

scripts/qesap/qesap.py
ricardobranco777: I dropped it because it's unused

mpagot: You are right, it is fine to drop

PR #57, opened 2022-09-02
Glue static test management

mpagot:


Makefile
```diff
@@ -2,8 +2,18 @@
...
 	@find -type f -name \*.sh -exec bash -n {} \;
+	@find -type f -name \*.sh -exec shellcheck {} -o all -s bash -S info \;
+
+static-py:
+	@find scripts -type f -name \*.py -exec pylint --rcfile=scripts/qesap/pylint.rc qesap.py {} +
 	@find scripts -type f -name \*.py -exec pylint --disable=line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,too-many-branches,unused-argument,too-many-locals,too-many-arguments,duplicate-code {} +
 	@find scripts -type f -name \*.py -exec flake8 --ignore=E501 {} +
+
+test:
 	@PYTHONPATH=scripts/qesap pytest-3
```
ricardobranco777, line 19: Maybe we should also change:

```suggestion
	@PYTHONPATH=scripts/qesap pytest
```

ricardobranco777 (approved):
I like the idea of having a pylintrc but for the whole repo so we can eventually cover also the Python files under other directories.

A small suggestion.  Otherwise LGTM

PR #58, opened 2022-09-02
Simplify subprocess.run

ricardobranco777:
Simplify subprocess.run to avoid having different code for different Python versions.


mpagot (approved):
LGTM

PR #59, opened 2022-09-02
Fix 2 return values

ricardobranco777:
This PR:
- Reuses main return value for boolean testing
- Creates a status class inheriting from `int` to attach status messages to integers interpreted as exit codes.

Both avoid the issue of having 2 return values and fixes the inconsistency in `cmd_ansible()` where sometimes it returns 2 values and sometimes 1.


mpagot (approved):
LGTM

ricardobranco777:
> LGTM

should we use `file=sys.stderr` in the place that I put the comment?

mpagot (approved):
LGTM, need a rebase

PR #61, opened 2022-09-05
Added wrapper to add weak idempotency to non-idempotent roles

mr-stringer:
This PR allows the HANA install and HANA System Replication playbooks to be run several times without changing the system.

Now the HANA install playbook and the HANA System Replication playbook touch files in /var/lib/qedep when run.  If this file is found by the playbook, running the role is skipped.

This means that the role won't fail, but it is not currently perfect, and it will prevent some operations in the future.  For example, if a user wants to manually uninstall HANA and run the automation again, the action will fail unless they know to remove the files in /var/lib/qedep.


alvarocarvajald (approved):
LGTM

Issue #64, opened 2022-09-06
[OpenQA][Azure] - test fails on ansible saptune setup


lpalovsky:
Test fails upon setting up saptune. Issue seems to happen because saptune systemd service is not existing on SUT. 

```
 'fatal: [vmhana01]: FAILED! => {', '    "changed": false,', '    "invocation": {', '        "module_args": {', '            "daemon_reexec": false,', '            "daemon_reload": false,', '            "enabled": false,', '            "force": null,', '            "masked": null,', '            "name": "saptune",', '            "no_block": false,', '            "scope": "system",', '            "state": "stopped"', '        }', '    },', '    "msg": "Could not find the requested service saptune: host"', '}', '', 'PLAY RECAP *********************************************************************', 'vmhana01                   : ok=5    changed=1    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0   ', 'vmhana02                   : ok=5    changed=1    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0   ', '']
DEBUG:QESAPDEP:Ansible process return ret:2
ERROR:QESAPDEP:command:['ansible-playbook', '-vvvv', '-i', '/root/qe_sap_deployment/terraform/azure/inventory.yaml', '/root/qe_sap_deployment/ansible/playbooks/sap-hana-preconfigure.yaml', '-e', 'use_sapconf=True'] return not zero 2
ZaIUs-2-
``` 
Test: https://mordor.suse.cz/tests/4384#
Log files: https://mordor.suse.cz/tests/4384#downloads

lpalovsky:
Problem might be in ansible checking for saptune sercive being disabled but fails since there is no saptune service present.

lpalovsky:
Problem seems to be in ansible checking for saptune service being disabled but fails since there is no saptune service present.

mr-stringer:
Which task is it failing on?  Is SAPCONF set to 'true' or 'false' when the runbook is called?

lpalovsky:
Sapconf is set to 'true' and it is failing at sap-hana-preconfigure.yaml
```
DEBUG:QESAPDEP:Ansible process return ret:2
ERROR:QESAPDEP:command:['ansible-playbook', '-vvvv', '-i', '/root/qe_sap_deployment/terraform/azure/inventory.yaml', '/root/qe_sap_deployment/ansible/playbooks/sap-hana-preconfigure.yaml', '-e', 'use_sapconf=True'] return not zero 2
[...]

PR #65, opened 2022-09-07
Update broken saptune package

lpalovsky:
Openqa test fails with missing saptune systemd service while package is installed.
Solution is to update package which will add systemd unit as well. Package is updated only if it already exists.

Details in ticket: https://github.com/SUSE/qe-sap-deployment/issues/64
VR:  https://mordor.suse.cz/tests/4396

Test continues past problematic part, fails later because another issue.


alvarocarvajald (approved):
LGTM

mr-stringer (approved):

mpagot (approved):
LGTM

PR #67, opened 2022-09-08
Env variable for Ansible

mpagot:
subprocess_run to also handle environment variables
set ANSIBLE_PIPELINING when running ansible-playbook

PR #72, opened 2022-09-15
Aws multiple sbd terraform

mr-stringer:
This PR adds support for AWS clusters using SBD fencing.

The terraform code had been altered in a similar way to the Azure terraform changed to enable multiple SBD servers there.

A lot of new AWS-specific ansible code has also been added.


ansible.cfg
```diff
@@ -1,2 +1,4 @@
 [ssh_connection]
 pipelining = True
+[defaults]
+executable = /bin/bash
```
mr-stringer, line 4: @mpagot For AWS the ansible configuration needs to include this setting.  It may need to be added to the glue code that same way as pipelining was!

mpagot: Could you achieve the same by setting some environment variable? Why is it an AWS specific one? Even if, strictly speaking only AWS need it, is it fine to run Azure and CGP with it too?

ansible.cfg
```diff
@@ -1,2 +1,4 @@
 [ssh_connection]
 pipelining = True
```
mpagot, line 2: Is it something that we can/must do to ensure that `ansible` and/or `ansible-playbook` are piking up this ansible.cfg file when they are called by the glue script? Is there an explicit command line argument to do it. If it is so we can use it and drop https://github.com/SUSE/qe-sap-deployment/commit/8a2cd7f25b2fd22dfaa4b8126efc4270ae65174d  at all

.gitignore
```diff
@@ -11,6 +11,7 @@ terraform/**/.terraform.lock.hcl
 terraform/**/.terraform.tfstate.lock.info
 terraform/**/inventory.yaml
 terraform/**/plan.zip
+terraform/aws/plan.zip/aws_cluster_data.yaml
```
mpagot, line 14: Usually plan.zip is a file not a folder

ansible/playbooks/cluster_sbd_prep.yaml
```diff
@@ -4,8 +4,12 @@
...
   become_user: root
   pre_tasks:
-    - name: include iscsi storage profile
-      ansible.builtin.include_vars: ./vars/iscsi-storage-profile.yaml
+    - name: Detect cloud platform
+      ansible.builtin.include_tasks:
+        ./tasks/detect-cloud-platform.yaml
+    - name: Detection result
+      ansible.builtin.debug:
+        msg: "Cloud platform appears to be {{ cloud_platform_name }}"
```
mpagot, line 12: is it done to have a better debug/log message? Is it more?

ansible/playbooks/pre-cluster.yaml
```diff
@@ -72,7 +79,17 @@
...
+    ansible.builtin.lineinfile:
+      path: /etc/cloud/cloud.cfg
+      regexp: '^preserve_hostname:'
+      line: 'preserve_hostname: true'
+      mode: '0544'
+      owner: root
+      group: root
+      create: yes
+      backup: yes
+    when: cloud_platform_is_aws
```
mpagot, line 92: Where `cloud_platform_is_aws` come from?

ansible/playbooks/cluster_sbd_prep.yaml
mr-stringer: For easier debugging.  Until this new task list is bedded in I decided to keep this on-screen so we could check that it was detecting the correct cloud platform.

.gitignore
mr-stringer: Ahh, good spot.  This is a typo, will fix.

ansible.cfg
mr-stringer: It's required for cloud detection as the command use a pipe to grep stdout from `dmidecode` to `grep`.  However, if it will be a problem I could look at getting rid of this and using ansible to do some regexp on stdout from `dmidecode` instead.

ansible.cfg
mr-stringer: I'm happy to set this in the glue code and get rid of `ansible.cfg` entirely.

ansible/playbooks/pre-cluster.yaml
mr-stringer: Yes.  detect os is run as a pre-task and that exposes its variables to this playbook.

ansible.cfg
mr-stringer: The above code has been removed.

.gitignore
mr-stringer: Fixed

ansible.cfg
mpagot: The ANSIBLE_PIPELINING is done at this line of PR#67 https://github.com/SUSE/qe-sap-deployment/pull/67/files#diff-bf288212e7712d3f535bcb117d5244f92dfc1318f8a7b8e70be61716743e2273R333

mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM. Confirmed also that changes to terraform code work still with my old tfvars file from #24

terraform/aws/modules/hana_node/main.tf
```diff
@@ -52,28 +52,37 @@ resource "aws_instance" "hana" {
...
+      device_name = ebs_block_device.value["device_name"]
+
+    }
   }
 
   volume_tags = {
     Name = "${var.common_variables["deployment_name"]}-${var.name}${format("%02d", count.index + 1)}"
   }
 
+  #maintenance_options {
```
mpagot, line 82: Why is has been committed as commented?

PR #73, opened 2022-09-20
Fix outputs for Azure

lpalovsky:
Currently azure outputs for hana nodes (ip addr, hostnames ... ) are not compatible with openqa requirements.
Data is structured inside nested lists which is not expected by openqa.
This PR makes the output same as previous ha-sap-terraform-deployment project.

Failed test output:
https://mordor.suse.cz/tests/4640#step/hana_sr_deployment/235

VR:
https://mordor.suse.cz/tests/4653


mpagot (approved):
LGTM

mr-stringer:
Looks good :)

PR #75, opened 2022-09-30
Gcp terraform multiple iscsi

mr-stringer:
This PR adds GCP support.

GCP is configurable with SBD only at this point.  1 to 3 iscsi servers may be requesrted.


terraform/gcp/inventory.tmpl
```diff
@@ -5,13 +5,15 @@ all:
 %{ for index, value in hana-pip ~}
         ${hana-name[index]}:
           ansible_host: ${value}
-          ansible_python_interpreter: /usr/bin/python3
+          %{ if hana-major-version == "12" }ansible_python_interpreter: /usr/bin/python2.7%{ else }ansible_python_interpreter: /usr/bin/python3 %{ endif }
```
mpagot, line 8: is this fix about the specific python version to use, specific to GCP?

lpalovsky (approved):

PR #76, opened 2022-10-04
Fixed AWS SLES 12 support

mr-stringer:


mpagot (approved):
LGTM (let me think that we have some code duplication but I do not have, for the moment, a better idea about how to avoid it)

mr-stringer:
> LGTM (let me think that we have some code duplication but I do not have, for the moment, a better idea about how to avoid it)

Yeah, we do, but it exists in different terraform directories for the different clouds.  We could try and merge some of the common stuff into a module.  But I think it's best left alone for the time being.

PR #88, opened 2022-10-17
Decrease ansible output verbosity

lpalovsky:
This PR decreases default ansible output verbosity while using 
--verbose cmd option. Adds option "verbosity" into yaml config file, 
if value is "true" output verbosity is increased to "-vvvv" 

VR:
https://mordor.suse.cz/tests/5147#

```
🐧 [lpalovsky@toolbox qesap/] PYTHONPATH=. pytest test/unit/
============================================================== test session starts ===============================================================
platform linux -- Python 3.10.4, pytest-6.2.4, py-1.11.0, pluggy-0.13.1
rootdir: /var/home/lpalovsky/GIT/qe-sap-deployment/scripts/qesap
collected 64 items                                                                                                                               

test/unit/test_qesap_ansible.py ............                                                                                               [ 18%]
test/unit/test_qesap_cli.py .............                                                                                                  [ 39%]
test/unit/test_qesap_configure.py ...............                                                                                          [ 62%]
test/unit/test_qesap_deploy.py .                                                                                                           [ 64%]
test/unit/test_qesap_destroy.py .                                                                                                          [ 65%]
test/unit/test_qesap_lib_config.py ...                                                                                                     [ 70%]
test/unit/test_qesap_terraform.py ..........s..                                                                                            [ 90%]
test/unit/test_subprocess_run.py ......                                                                                                    [100%]

========================================================= 63 passed, 1 skipped in 0.31s ==========================================================

```

### DIFF
--- a/scripts/qesap/test/unit/test_qesap_ansible.py
+++ b/scripts/qesap/test/unit/test_qesap_ansible.py
@@ -44,9 +44,9 @@ def test_ansible_create(run, _, base_args, tmpdir, create_inventory, create_play
 
 @mock.patch('shutil.which', side_effect = [(ANSIBLEPB_EXE),(ANSIBLE_EXE)])
 @mock.patch("lib.process_manager.subprocess_run")
-def test_ansible_verbose(run, _, base_args, tmpdir, create_inventory, create_playbooks, ansible_config, mock_call_ansibleplaybook):
+def test_ansible_verbose_cmd(run, _, base_args, tmpdir, create_inventory, create_playbooks, ansible_config, mock_call_ansibleplaybook):
     """
-    run with -vvvv if qesap ansible --verbose
+    running - 'qesap ansible --verbose' should not trigger '-vvvv'.
     (probably not supported in qesap deploy/destroy)
     """
     provider = 'grilloparlante'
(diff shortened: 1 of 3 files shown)


scripts/qesap/test/unit/test_qesap_ansible.py
```diff
@@ -50,8 +50,9 @@ def test_ansible_verbose(run, _, base_args, tmpdir, create_inventory, create_pla
     """
     provider = 'grilloparlante'
     playbooks = {'create': ['get_cherry_wood']}
+    verbosity = "true"
```
mpagot, line 53: why not to replace 
```
 verbosity = "true"

    config_content = ansible_config(provider, playbooks, verbosity)
    ```
    
    with 
    ```
    config_content = ansible_config(provider, playbooks, "true")
```

scripts/qesap/test/unit/test_qesap_ansible.py
```diff
@@ -50,8 +50,9 @@ def test_ansible_verbose(run, _, base_args, tmpdir, create_inventory, create_pla
...
 
-    config_content = ansible_config(provider, playbooks)
+    config_content = ansible_config(provider, playbooks, verbosity)
```
mpagot, line 55: This test is ok, but I propose you to add one more test, not using *ansible_config* and composing, directly in the test function, a config.yaml with no *verbosity:* at all (and not only *verbosity:false* )

mpagot: I mean, what is the expected qesap.py behavior if a user omit *verbosity:* at all in its config.yaml? Maybe create one more test about this scenario.

mpagot: I just realize that you adapt an existing test but my suggestion is to create two more tests focused on the presence or absence of -vvvv in the generated ansible command, one test about starting with a config.yaml with false, another one without and verbosity key at all in the config.yaml.  You may also test that a script executed with *qesap.py --verbose* and *verbosity: false* in the config.yaml, result in Ansible called without -vvvv

mpagot (approved):
LGTM

PR #94, opened 2022-10-24
Ansible to run on the original env

mpagot:
Ansible executed on the original env with the addition (and not substitution) of ANSIBLE_PIPELINING.

### DIFF
--- a/scripts/qesap/test/conftest.py
+++ b/scripts/qesap/test/conftest.py
@@ -320,6 +320,8 @@ def mock_call_ansibleplaybook():
     ```
     '''
     def _callback(playbook_cmd):
-        return mock.call(cmd=playbook_cmd, env={'ANSIBLE_PIPELINING': 'True'})
+        original_env = dict(os.environ)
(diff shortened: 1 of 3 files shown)


scripts/qesap/test/conftest.py
```diff
@@ -320,6 +320,8 @@ def mock_call_ansibleplaybook():
     ```
     '''
     def _callback(playbook_cmd):
-        return mock.call(cmd=playbook_cmd, env={'ANSIBLE_PIPELINING': 'True'})
+        original_env = dict(os.environ)
```
ricardobranco777, line 323: This doesn't create a copy of the `os.environ` dictionary.  

```suggestion
        original_env = os.environ.copy()
```

ricardobranco777 (changes requested):
The code assiging the `os.environ` dict doesn't make a copy as `original_env`.  You must use the `.copy()` method instead.

ricardobranco777 (approved):
LGTM

PR #106, opened 2022-12-02
Fail when there are no repos at the end

dzedro:
-don't check if registercloudguest is present when there are repos
-keep order of when conditon, rcg.rc can't be before repos.rc because it will fail as it won't have value to compare
-cleanup white spaces

The issues I had were because different os_image was used than I expected due to mess which will be fixed in https://jira.suse.com/browse/TEAM-6941 & https://jira.suse.com/browse/TEAM-6950
I just check at the end that there are repos, the problems I had with registercloudguest will probably not happen.
PAYG images will probably need sleep or retry at `Check for registration`, because this images add repos but it takes some time, without retry we check before the repos are added. Should I add it now ?
```
     - name: Check for registration
       ansible.builtin.command: zypper lr
       register: repos
       failed_when: false
       changed_when: false
+      retries: 3
+      delay: 60
```

VR:
https://openqaworker15.qa.suse.cz/tests/96420
https://openqaworker15.qa.suse.cz/tests/96421

### DIFF
--- a/ansible/playbooks/registration.yaml
+++ b/ansible/playbooks/registration.yaml
@@ -8,7 +8,7 @@
 
     #Pre flight checks
 
-    #Do we have repos?  If yes, we need to register
+    #Do we have repos?  If not, we need to register
     - name: Check for registration
       ansible.builtin.command: zypper lr
       register: repos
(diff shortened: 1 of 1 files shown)


mr-stringer (approved):
This looks good.  When I did this a role recently I took a slightly different approach.  I used SUSEConnect output to see if the base product was registered as it provides JSON output so is a little more predictable.  However, this approach seems good too.

https://github.com/mr-stringer/sles-for-sap-roles/blob/main/roles/sles_register/tasks/post-checks.yml

PR #115, opened 2023-01-04
Disable loadbalancer on GCP for non-HA scenarios

lpalovsky:
Load balancer setup is executed by default on GCP regardless of ha being 
enabled. This PR disables load balancer outside of ha scenario and single node 
deployment.

VR: 
single node: https://mordor.suse.cz/tests/5780
cluster:  https://mordor.suse.cz/tests/5778/  (Issue not relevant to terraform)
mr_test (originally affected, fail unrelated):  http://openqaworker15.qa.suse.cz/tests/121245/#

### DIFF
--- a/terraform/gcp/modules/hana_node/main.tf
+++ b/terraform/gcp/modules/hana_node/main.tf
@@ -5,6 +5,7 @@ locals {
   create_ha_infra        = var.hana_count > 1 && var.common_variables["hana"]["ha_enabled"] ? 1 : 0
   provisioning_addresses = google_compute_instance.clusternodes.*.network_interface.0.access_config.0.nat_ip
   hostname               = var.common_variables["deployment_name_in_hostname"] ? format("%s-%s", var.common_variables["deployment_name"], var.name) : var.name
+  balancer_groups        = var.hana_count > 1 && var.common_variables["hana"]["ha_enabled"] ? ["hana-primary-group", "hana-secondary-group"] : []
 }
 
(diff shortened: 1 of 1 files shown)


dzedro (approved):
I don't understand tf that much, but code make sense and VR passed. :+1:

lpalovsky:
> I don't understand tf that much, but code make sense and VR passed. +1

The issue was that original code would create load balancers even if there was only one host being deployed. Afterwards it would try to assign the non existing second host to the secondary balancer an failing.

mr-stringer (approved):
LGTM.

I ran the code with and without HA to check that the code behaves as expected. It does. However, the Ansible for creating the volumes is still broken since the deployment name was added to the disk name. It's not related to this change, but it still means that the overall deployment doesn't work.

lilyeyes (approved):
I don't familiar with this project (https://github.com/SUSE/qe-sap-deployment) but the code changes and VRs really lgtm.

lpalovsky:
> LGTM.
> 
> I ran the code with and without HA to check that the code behaves as expected. It does. However, the Ansible for creating the volumes is still broken since the deployment name was added to the disk name. It's not related to this change, but it still means that the overall deployment doesn't work.

Yes I am aware about that one. Haven't reached that point yet, but will be there soon. I think Michelle was dealing with it so we probably have a Jira ticket in queue already.

mpagot (approved):
LGTM

PR #123, opened 2023-02-20
Move --workspace as a sub argument to terraform command

alvarocarvajald:
Move `--workspace` as a sub argument to terraform command.

This is a followup to #122 

Verification Runs: [HanaSR Regression Test-15-SP3@Azure](https://openqaworker15.qa.suse.cz/tests/129438), [saptune test-15-SP4@Azure](http://openqaworker15.qa.suse.cz/tests/129437), [saptune test-12-SP4@EC2](http://openqaworker15.qa.suse.cz/tests/129439)

PR #124, opened 2023-02-21
Introduce os_image_uri Terraform variable in Azure

mpagot:
Add a new Terraform variable `os_image_uri` that is similar to `os_image` but for custom images. It is on top of module specific _uri variables.
Change the code to allow to run an Azure deployment only with _uri variables and without any os_image

### DIFF
--- a/terraform/azure/main.tf
+++ b/terraform/azure/main.tf
@@ -50,6 +50,12 @@ locals {
   drbd_os_image       = var.drbd_os_image != "" ? var.drbd_os_image : var.os_image
   netweaver_os_image  = var.netweaver_os_image != "" ? var.netweaver_os_image : var.os_image
 
+  hana_os_image_uri       = var.sles4sap_uri != "" ? var.sles4sap_uri : var.os_image_uri
+  iscsi_os_image_uri      = var.iscsi_srv_uri != "" ? var.iscsi_srv_uri : var.os_image_uri
+  monitoring_os_image_uri = var.monitoring_uri != "" ? var.monitoring_uri : var.os_image_uri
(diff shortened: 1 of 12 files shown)


alvarocarvajald (approved):
LGTM

alvarocarvajald:
We would need to certify after merging that working tests are deploying the cloud instances the OS we're specifying, specially in those cases where both `os_image` and `sles4sap_uri` are defined in the tfvars.

lilyeyes:
In https://openqaworker15.qa.suse.cz/tests/132731#step/qesap_terraform/257 it failed on "Terraform deployment FAILED"

lpalovsky (approved):
LGTM

jankohoutek:
I would name it `os_custom_image_uri` so it will be not potentially confusing

mpagot:
> In https://openqaworker15.qa.suse.cz/tests/132731#step/qesap_terraform/257 it failed on "Terraform deployment FAILED"

Yes, error is https://openqaworker15.qa.suse.cz/tests/132731#step/qesap_terraform/248

I think that my PR is not including https://github.com/SUSE/qe-sap-deployment/pull/123

mpagot:
> I would name it `os_custom_image_uri` so it will be not potentially confusing

Thanks, if it is fine for you I'll implement it in a next PR that I need in any case to create to implement some more variable validations

lilyeyes:
> In https://openqaworker15.qa.suse.cz/tests/132731#step/qesap_terraform/257 it failed on "Terraform deployment FAILED"

Thanks, "Terraform deployment" passed now. I have no more comments as I am not familiar with this project.

PR #134, opened 2023-03-29
Give standard name to Hana VMs

mpagot:
Use same name used in AWS and Azure deployment like `vmhana01`.

Ticket [TEAM-7627](https://jira.suse.com/browse/TEAM-7627)

Verification: http://openqaworker15.qa.suse.cz/tests/141285 http://openqaworker15.qa.suse.cz/tests/141286


mpagot:
Not the right approach

PR #142, opened 2023-04-26
Add no Ansible support

mpagot:
Allow qesap to run without Ansible.
Re-organize the cmd ansible, extend CONF class with more API about ansible validation.


alvarocarvajald (approved):
LGTM

PR #148, opened 2023-05-09
Add two targets to run KICS

mpagot:
Add two optional targets to the main Makefile to run KICS on Ansible and Terraform separately. Add a configuration file and some error suppression about public IPs.

Ticket: [TEAM-7231](https://jira.suse.com/browse/TEAM-7231)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 # Flake8 codes:
 # E501 line too long (XX > 79 characters)
 
+base_dir := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
 all: static test
 
 static: static-bash static-py static-ansible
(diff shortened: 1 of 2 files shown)


alvarocarvajald:
Is this run as part of the merge pipeline?

mpagot:
> Is this run as part of the merge pipeline?
No it is not. Even in the Makefile I explicitly decided not to include it in the `all` target. The idea of this PR is to publish some code to solve the problem about how to run the tool. The two suppression that I added are about PublicIp warning. We know that we have public IP and it is on purpose.

jankohoutek (commented):
LGTM

jankohoutek:
> We know that we have public IP and it is on purpose.

Yes, but do we does it  correctly? 
I mean, do we follow the recommended additional security baselines on Azure? 
https://docs.microsoft.com/en-us/security/benchmark/azure/baselines/public-ip-security-baseline

PR #150, opened 2023-05-10
Prevent deploying with "default" deployment name

lilyeyes:
TEAM-7850 - Prevent qe-sap-deployment from deploying with the default deployment name


mpagot (approved):
LGTM

PR #155, opened 2023-06-03
Check hana_media not to be full ulr

mpagot:
Return an explicit error message during `qesap.py configure` if the user provide `hana_media` as a full URL even when apiver:3 is used. The expected value is a filename, Ansible will compose the full url from it.

Verification:
 - http://openqaworker15.qa.suse.cz/tests/185512
 - http://openqaworker15.qa.suse.cz/tests/185511 intentionally triggered with wrong `QESAPDEPLOY_SAPCAR` setting. Here the improved error message http://openqaworker15.qa.suse.cz/tests/185511#step/configure/155 . Now it fails directly during configure and not later during Ansible.


BillAnastasiadis (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #159, opened 2023-06-16
Change way to detect unregistered nodes

BillAnastasiadis:
Our ansible `registration` playbook relies on `zypper lr` returning no repos to detect whether registration is needed. This does not help in the case of maintenance repos however, as they have to be added before ansible, which leads to the playbook wrongly acting like no registration is needed.

This pr addresses that by using `SUSEConnect -s` instead.

- Related Ticket: https://jira.suse.com/browse/TEAM-8075
- Verification Run: http://openqaworker15.qa.suse.cz/tests/190396
and a PAYG aws one: http://openqaworker15.qa.suse.cz/tests/190397

### DIFF
--- a/ansible/playbooks/registration.yaml
+++ b/ansible/playbooks/registration.yaml
@@ -10,10 +10,16 @@
 
     #Do we have repos?  If not, we need to register
     - name: Check for registration
-      ansible.builtin.command: zypper lr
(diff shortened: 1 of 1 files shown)


mpagot (approved):
LGTM

rjagu1:
LGTM

rjagu1 (approved):

PR #153, opened 2023-05-18
Make reboot timeout configurable

rjagu1:
Increase Timeout from Default 600 to 900 to overcome the reboot timeout failures in deployments   
REBOOT_TIMEOUT: 900
 - preconfigure.yaml -e use_sapconf=${SAPCONF} -e use_reboottimeout=${REBOOT_TIMEOUT}

Ticket[TEAM-6829](https://jira.suse.com/browse/TEAM-6829)

VR link on Azure: http://openqaworker15.qa.suse.cz/tests/179777
http://openqaworker15.qa.suse.cz/tests/179786

PR #161, opened 2023-06-29
Fix resource ip not starting

BillAnastasiadis:
This pr makes sure `rsc_ip` is created dynamically together with the hana ips. 
In case the user provides custom ips manually, they should make sure to provide a suitable `rsc_ip` as well.

- Related ticket: https://jira.suse.com/browse/TEAM-8069
- Verification Run: http://openqaworker15.qa.suse.cz/tests/195928#step/deploy_qesap_ansible/85

### DIFF
--- a/ansible/playbooks/tasks/azure-cluster-hana.yaml
+++ b/ansible/playbooks/tasks/azure-cluster-hana.yaml
@@ -60,7 +60,7 @@
 
 - name: Create IP resource
   # TODO - IP address is hardcoded here, that's not good and will only work when the VIP perfectly matches.
-  ansible.builtin.command: "crm configure primitive rsc_ip_{{ sap_hana_install_sid }}{{ sap_hana_install_instance_number }} ocf:heartbeat:IPaddr2 meta target-role=\"Started\" operations $id=\"rsc_ip_HDB_{{ sap_hana_install_sid }}{{ sap_hana_install_instance_number }}-operations\" op monitor interval=\"10s\" timeout=\"20s\" params ip=\"10.10.1.12\""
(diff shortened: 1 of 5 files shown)


alvarocarvajald (approved):
LGTM. Only issue I have is that the `rsc_ip` variable name does not immediately informs what it's supposed to define. Is it possible to rename it to something like `cluster_ip` or `cluster_alias_ip`? If you need more inspiration about what the `ocf:heartbeat:IPaddr2` resource type implements, you can see here: https://www.mankier.com/7/ocf_heartbeat_IPaddr2

mpagot:
As Terraform and Ansible should maintain a sort of autonomy one to the other, maybe it make sense to define a default on the top of ansible/playbooks/tasks/azure-cluster-hana.yaml (something like https://github.com/SUSE/qe-sap-deployment/pull/153/files#diff-9fc1971cca7161106f3df15e6caf44425c11a96264c6acc9a8a9a854eb07143eR10) or to to define a task to validate the variable content. I'm thinking to prompt the user with a clear error in case a deployment is missing the new variable in the inventory. 
It is just a proposal and I'm not sure if it really makes sense.

mpagot (approved):
LGTM

BillAnastasiadis:
> LGTM. Only issue I have is that the `rsc_ip` variable name does not immediately informs what it's supposed to define. Is it possible to rename it to something like `cluster_ip` or `cluster_alias_ip`? If you need more inspiration about what the `ocf:heartbeat:IPaddr2` resource type implements, you can see here: https://www.mankier.com/7/ocf_heartbeat_IPaddr2

thanks Alvaro, done! New VR is running, I'll replace the current one

PR #176, opened 2023-09-22
Replace azurerm vm with azurerm_linux

BillAnastasiadis:
This pr replaces all cases of `azurerm_virtual_machine`, which is deprecated, with `azurerm_linux_virtual_machine`. The conversion demanded a LOT of other changes as well, as the 2 resources have a lot of differences in the way they define and use their components.

**Some of the differences (they are too many to list all of them) are:**

- Data disks that were defined inside the `azurerm_virtual_machine` resource, now need to be defined as a separate resource together with an attachment resource that binds them to the respective VMs.
- That also lead to changes in the way the names of the disks are calculated, since some of them used VM specific counts. I did the changes with the goal of ending up with the same exact names in the end.
- `vm_size` renamed to `size`
- The `os_profile` and `os_profile_linux_config` blocks are replaced with top-level arguments in `azurerm_linux_virtual_machine`
- `delete_os_disk_on_termination` and `delete_data_disks_on_termination` attributes are removed from the `azurerm_linux_virtual_machine` resource. The first is now moved to provider level and renamed (`provider->features->delete_os_disk_on_deletion`), while the second has disappeared completely (!!) and is no longer doable in terraform. Ricardo said thet pcw takes care of that cleanup
- `boot_diagnostics` changes: `enabled` attribute no longer supported, while `storage_uri` is renamed to `storage_account_uri`
- ... and more.

**LINKS:**
- Related ticket: https://jira.suse.com/browse/TEAM-8424
- Verification Runs: 

**HanaSr-Azure-(BYOS|PAYG):**
15sp3 BYOS:
https://openqaworker15.qa.suse.cz/tests/236441 :green_circle: 

**Qesap-Azure-(BYOS|PAYG)** (failure after deployment expected):
15sp5:
https://openqaworker15.qa.suse.cz/tests/236437 red is later after the deployment and quite like due to TEAM-8442
15sp4:
https://openqaworker15.qa.suse.cz/tests/236439 red is later after the deployment and quite like due to TEAM-8442
12sp5:
https://openqaworker15.qa.suse.cz/tests/237155 :green_circle: 
http://openqaworker15.qa.suse.cz/tests/237156 red is later after the deployment and quite like due to TEAM-8442

**qesap-azure-saptune-uri:**
https://openqa.suse.de/tests/12325535 red is later after the deployment and quite like due to TEAM-8442

### DIFF
--- a/terraform/azure/modules/hana_node/main.tf
+++ b/terraform/azure/modules/hana_node/main.tf
@@ -364,62 +364,60 @@ locals {
   disks_writeaccelerator = [for writeaccelerator in split(",", var.hana_data_disks_configuration["writeaccelerator"]) : tobool(trimspace(writeaccelerator))]
 }
 
-resource "azurerm_virtual_machine" "hana" {
-  count                            = var.hana_count
-  name                             = "${var.name}${format("%02d", count.index + 1)}"
-  location                         = var.az_region
-  resource_group_name              = var.resource_group_name
-  network_interface_ids            = [element(azurerm_network_interface.hana.*.id, count.index)]
-  availability_set_id              = var.common_variables["hana"]["ha_enabled"] ? azurerm_availability_set.hana-availability-set[0].id : null
-  vm_size                          = var.vm_size
-  delete_os_disk_on_termination    = true
-  delete_data_disks_on_termination = true
+resource "azurerm_managed_disk" "hana_data_disk" {
+  count                = var.hana_count * local.disks_number
+  name                 = "disk-${var.name}${format("%02d", floor(count.index / local.disks_number) + 1)}-Data${format("%02d", count.index % local.disks_number + 1)}"
+  location             = var.az_region
+  resource_group_name  = var.resource_group_name
+  storage_account_type = element(local.disks_type, count.index % local.disks_number)
+  create_option        = "Empty"
+  disk_size_gb         = element(local.disks_size, count.index % local.disks_number)
+}
 
-  storage_os_disk {
-    name              = "disk-${var.name}${format("%02d", count.index + 1)}-Os"
-    caching           = "ReadWrite"
-    create_option     = "FromImage"
-    managed_disk_type = "Premium_LRS"
+resource "azurerm_virtual_machine_data_disk_attachment" "hana_data_disk_attachment" {
+  count              = var.hana_count * local.disks_number
+  managed_disk_id    = azurerm_managed_disk.hana_data_disk[count.index].id
+  virtual_machine_id = azurerm_linux_virtual_machine.hana[floor(count.index / local.disks_number)].id
+  lun                = count.index % local.disks_number
+  caching            = element(local.disks_caching, count.index % local.disks_number)
+}
+
+resource "azurerm_linux_virtual_machine" "hana" {
+  count                 = var.hana_count
+  name                  = "${var.name}${format("%02d", count.index + 1)}"
+  location              = var.az_region
+  resource_group_name   = var.resource_group_name
+  network_interface_ids = [element(azurerm_network_interface.hana.*.id, count.index)]
+  availability_set_id   = var.common_variables["hana"]["ha_enabled"] ? azurerm_availability_set.hana-availability-set[0].id : null
+  size                  = var.vm_size
+
+  admin_username = var.common_variables["authorized_user"]
+  admin_ssh_key {
+    username   = var.common_variables["authorized_user"]
+    public_key = var.common_variables["public_key"]
   }
+  disable_password_authentication = true
 
-  storage_image_reference {
-    id        = var.sles4sap_uri != "" ? join(",", azurerm_image.sles4sap.*.id) : ""
-    publisher = var.sles4sap_uri != "" ? "" : module.os_image_reference.publisher
-    offer     = var.sles4sap_uri != "" ? "" : module.os_image_reference.offer
-    sku       = var.sles4sap_uri != "" ? "" : module.os_image_reference.sku
-    version   = var.sles4sap_uri != "" ? "" : module.os_image_reference.version
-  }
-
-  dynamic "storage_data_disk" {
-    for_each = [for v in range(local.disks_number) : { index = v }]
+  dynamic "source_image_reference" {
+    for_each = var.sles4sap_uri != "" ? [] : [1]
     content {
-      name                      = "disk-${var.name}${format("%02d", count.index + 1)}-Data${format("%02d", storage_data_disk.value.index + 1)}"
-      managed_disk_type         = element(local.disks_type, storage_data_disk.value.index)
-      create_option             = "Empty"
-      lun                       = storage_data_disk.value.index
-      disk_size_gb              = element(local.disks_size, storage_data_disk.value.index)
-      caching                   = element(local.disks_caching, storage_data_disk.value.index)
-      write_accelerator_enabled = element(local.disks_writeaccelerator, storage_data_disk.value.index)
+      publisher = module.os_image_reference.publisher
+      offer     = module.os_image_reference.offer
+      sku       = module.os_image_reference.sku
+      version   = module.os_image_reference.version
     }
   }
 
-  os_profile {
-    computer_name  = "${local.hostname}${format("%02d", count.index + 1)}"
-    admin_username = var.common_variables["authorized_user"]
-  }
+  source_image_id = var.sles4sap_uri != "" ? join(",", azurerm_image.sles4sap.*.id) : null
 
-  os_profile_linux_config {
-    disable_password_authentication = true
-
-    ssh_keys {
-      path     = "/home/${var.common_variables["authorized_user"]}/.ssh/authorized_keys"
-      key_data = var.common_variables["public_key"]
-    }
+  os_disk {
+    name                 = "disk-${var.name}${format("%02d", count.index + 1)}-Os"
+    caching              = "ReadWrite"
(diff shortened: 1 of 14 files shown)


terraform/azure/modules/hana_node/main.tf
```diff
@@ -364,62 +364,60 @@ locals {
...
-
-  storage_os_disk {
-    name              = "disk-${var.name}${format("%02d", count.index + 1)}-Os"
-    caching           = "ReadWrite"
-    create_option     = "FromImage"
-    managed_disk_type = "Premium_LRS"
-  }
+resource "azurerm_managed_disk" "hana_data_disk" {
+  count                = var.hana_count * local.disks_number
+  name                 = "disk-${var.name}${format("%02d", floor(count.index / local.disks_number) + 1)}-Data${format("%02d", count.index % local.disks_number + 1)}"
```
mpagot, line 369: Was
```
name = "${var.name}${format("%02d", count.index + 1)}"
```

why a more complicated disk name is needed?

terraform/azure/modules/hana_node/main.tf
```diff
@@ -364,62 +364,60 @@ locals {
...
-  delete_data_disks_on_termination = true
-
-  storage_os_disk {
-    name              = "disk-${var.name}${format("%02d", count.index + 1)}-Os"
-    caching           = "ReadWrite"
-    create_option     = "FromImage"
-    managed_disk_type = "Premium_LRS"
-  }
+resource "azurerm_managed_disk" "hana_data_disk" {
+  count                = var.hana_count * local.disks_number
```
mpagot, line 368: Why `local.disks_number` is needed in the new `count` formula?

terraform/azure/modules/hana_node/main.tf
```diff
@@ -364,62 +364,60 @@ locals {
...
+}
 
-  storage_image_reference {
-    id        = var.sles4sap_uri != "" ? join(",", azurerm_image.sles4sap.*.id) : ""
-    publisher = var.sles4sap_uri != "" ? "" : module.os_image_reference.publisher
-    offer     = var.sles4sap_uri != "" ? "" : module.os_image_reference.offer
-    sku       = var.sles4sap_uri != "" ? "" : module.os_image_reference.sku
-    version   = var.sles4sap_uri != "" ? "" : module.os_image_reference.version
+resource "azurerm_virtual_machine_data_disk_attachment" "hana_data_disk_attachment" {
+  count              = var.hana_count * local.disks_number
```
mpagot, line 378: same as before about `count` formula

terraform/azure/modules/hana_node/main.tf
```diff
@@ -364,62 +364,60 @@ locals {
...
+  count                 = var.hana_count
+  name                  = "${var.name}${format("%02d", count.index + 1)}"
+  location              = var.az_region
+  resource_group_name   = var.resource_group_name
+  network_interface_ids = [element(azurerm_network_interface.hana.*.id, count.index)]
+  availability_set_id   = var.common_variables["hana"]["ha_enabled"] ? azurerm_availability_set.hana-availability-set[0].id : null
+  size                  = var.vm_size
+
+  admin_username = var.common_variables["authorized_user"]
+  admin_ssh_key {
```
mpagot, line 395: is no more 
```
path = "/home/${var.common_variables["authorized_user"]}/.ssh/authorized_keys"
```

needed or supported?

terraform/azure/modules/majority_maker_node/main.tf
```diff
@@ -64,48 +64,43 @@ module "os_image_reference" {
...
+    for_each = var.sles4sap_uri != "" ? [] : [1]
+    content {
+      publisher = module.os_image_reference.publisher
+      offer     = module.os_image_reference.offer
+      sku       = module.os_image_reference.sku
+      version   = module.os_image_reference.version
     }
   }
 
+  source_image_id = var.sles4sap_uri != "" ? join(",", azurerm_image.sles4sap.*.id) : null
```
mpagot, line 100: Why this rework was needed?

```
 dynamic "source_image_reference" {
    for_each = var.sles4sap_uri != "" ? [] : [1]
    content {
      publisher = module.os_image_reference.publisher
      offer     = module.os_image_reference.offer
      sku       = module.os_image_reference.sku
      version   = module.os_image_reference.version
    }
  }

  source_image_id = var.sles4sap_uri != "" ? join(",", azurerm_image.sles4sap.*.id) : null
```
This deserve a dedicated VR

terraform/azure/modules/monitoring/main.tf
```diff
@@ -64,57 +64,59 @@ module "os_image_reference" {
...
-
-  storage_os_disk {
-    name              = "disk-monitoring-Os"
-    caching           = "ReadWrite"
-    create_option     = "FromImage"
-    managed_disk_type = "Premium_LRS"
-  }
+resource "azurerm_managed_disk" "monitoring_data" {
+  count                = var.monitoring_enabled == true ? 1 : 0
+  name                 = "disk-monitoring-Data01"
```
mpagot, line 69: In VM, the name is controllable by a variable before

https://github.com/SUSE/qe-sap-deployment/blob/c35d2805b4697d78916e4d6c6d061ea3671af345/terraform/azure/modules/monitoring/variables.tf#L20

Does not make sense to attach the name here too? Maybe not?

mpagot (commented):
No changes to propose but some questions and one more VR to run

lpalovsky (approved):
LGTM

terraform/azure/modules/hana_node/main.tf
BillAnastasiadis: It's not needed OR supported. From the documentation:

NOTE:
The Azure VM Agent only allows creating SSH Keys at the path /home/{username}/.ssh/authorized_keys - as such this public key will be written to the authorized keys file.

terraform/azure/modules/hana_node/main.tf
BillAnastasiadis: because the name needs to be unique for each disk, and since before we were using a loop for each disk for each VM, now that the loop is gone we have to incorporate it in the variables (count is used to let terraform know how many resources to spawn: https://buildvirtual.net/terraform-count-examples/). Also see answer in the next question

terraform/azure/modules/majority_maker_node/main.tf
BillAnastasiadis: Because `source_image_id` was no longer supported inside the `dynamic "source_image_reference"`, which replaces `storage_image_reference`. Now there's `source_image_reference` for catalogue images, and `source_image_id` for custome images, but documentation states

NOTE:
One of either source_image_id or source_image_reference must be set.

I resorted to that code that checks for the existance of `var.sles4sap_uri` and uses one or the other accordingly. It's basically the same logic as the existing code, just with the id separated as a different var outside the reference as the new specification states

terraform/azure/modules/monitoring/main.tf
BillAnastasiadis: This is the data disk, and it had a set name from before - I just kept it as is:

```
storage_data_disk {
    name              = "disk-monitoring-Data01"
    caching           = "ReadWrite"
    create_option     = "Empty"
    disk_size_gb      = "10"
    lun               = "0"
    managed_disk_type = "Standard_LRS"
```

terraform/azure/modules/majority_maker_node/main.tf
alvarocarvajald: Thanks for the explanation Bill. I guess that means we need a VR for each method the image can be supplied, either via an URI which would mean we are testing an unpublished image from the Build Service, or by submitting a publisher, version, sku, etc, meaning we are testing on a catalog image.

alvarocarvajald (approved):
LGTM, but I think we need to add more verification runs to cover more the changes.

mpagot (approved):
LGTM

BillAnastasiadis:
> LGTM, but I think we need to add more verification runs to cover more the changes.

@alvarocarvajald suggest away! I'm running another one based on https://openqa.suse.de/tests/12289218, feel free to ask for others as well

alvarocarvajald:
> > LGTM, but I think we need to add more verification runs to cover more the changes.
> 
> @alvarocarvajald suggest away! I'm running another one based on https://openqa.suse.de/tests/12289218, feel free to ask for others as well

Off the top of my head, I'd say some from https://openqa.suse.de/group_overview/285 (HanaSR on PAYG and BYOS, saptune on PAYG & BYOS) which deal with unreleased OS images, and then perhaps some from https://openqaworker15.qa.suse.cz/group_overview/41 (I'd say 15-SP5 and 12-SP5, since you already have a passing one with 15-SP3). Finally some saptune ones from https://openqaworker15.qa.suse.cz/group_overview/46 (saptune on 15-SP5 PAYG mostly, since the previous 15-SP5 ones are BYOS) and the same from https://openqaworker15.qa.suse.cz/group_overview/28 (saptune on 12-SP5 ... but last MU which triggered tests in this job group is 18 days old, so leave this last one as optional)

a-kpappas:
@BillAnastasiadis Some VRs failed. Do you know why?

BillAnastasiadis:
> @BillAnastasiadis Some VRs failed. Do you know why?
> 
> (The operation was succesful. The patient died.)

@a-kpappas someone edited the description and answered :D

a-kpappas:
LGTM

PR #189, opened 2023-11-13
Move Ansible modules dependency to Galaxy

mpagot:
Only keep Ansible-Core as OS dependency and move all needed Ansible modules installation to use ansible-galaxy.

It is more aligned as modern Ansible best practice 

Ticket  TEAM-8786

# Verification

## qesap regression

http://openqaworker15.qa.suse.cz/tests/267459 :green_circle: 

Galaxy installation has its own dedicated log http://openqaworker15.qa.suse.cz/tests/267459/logfile?filename=configure-galaxy_install.txt

To have an idea about the overall installation process you need to look at http://openqaworker15.qa.suse.cz/tests/267459/logfile?filename=serial_terminal.txt and look for `pip3.10 install --no-color --no-cache-dir -r` and `ansible-galaxy install -r` 

## hanasr

http://openqaworker15.qa.suse.cz/tests/267460  :green_circle:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,3 @@
-ansible==6.5.0
-ansible-compat==2.2.1
 ansible-core==2.13.5
 awscli==1.29.4
 jmespath==0.10.0
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Tools needed
 
 * Python 3.9
 * terraform v1.3.6
-* ansible 6.5.0 (ansible-core 2.13.5). Please refer to the **requirements.txt** file
+* ansible-core 2.13.5 : please refer to the **requirements.txt** file
 * cloud provider cli tools (`az`, `aws`, `gcloud`)
 
 The Python requirements could be managed with a virtual environment
@@ -19,6 +19,12 @@ The Python requirements could be managed with a virtual environment
 python3 -m venv venv
 source venv/bin/activate
 pip install -r requirements.txt
+```
+
+Ansible dependency are managed separately. After the installation of the Ansible core the following command has to be executed:
+
+```shell
+ansible-galaxy install -r requirements.yml
 ```
 
 Prepare a ssh key pair
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,0 +1,8 @@
+---
+collections:
+  - name: community.crypto
+    version: 2.15.1
+  - name: ansible.posix
+    version: 1.5.4
+  - name: community.general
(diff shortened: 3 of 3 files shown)


lpalovsky (approved):
LGTM

mpagot:
> 1. JIRA?

Here a dedicated ticket just created for this topic TEAM-8786 

>     2. Would running this on a TW container and using the ansible package not provide everything needed?

Are you suggesting not installing Ansible with `pip install ansible-core` but using something like `zypper in ansible` ?
 
>     3. If not, why not package them that way?

Which kind of package are you referring to? `zypper` package?

requirements.txt
```diff
@@ -1,5 +1,3 @@
-ansible==6.5.0
-ansible-compat==2.2.1
```
alvarocarvajald, line 2: These will not break existing tests because https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/18133 is already merged, right? Just to confirm.

alvarocarvajald (approved):
LGTM, although I'm no expert. I have a question though.

requirements.txt
mpagot: yes, the fact that PR#18133 has been merged in advance result in test not finding the requirements.yaml (as this PR is not already in, or if test are using fixed old version of qe-sap-deployment) and so installing  `ansible==6.5.0` and having everything installed as `ansible==6.5.0` is the old approach with everything inside it.

Once committing this PR, the test will initially not installing `ansible==6.5.0` but the smaller `ansible-core` using pip. But as this same PR also has a new file named `requirements.yaml`, this will result on the test side in `ansible-galaxy install` to be called too, that will result in installing everything that Ansible needs and was previously provided by `ansible==6.5.0`.

lpalovsky (approved):
LGTM

a-kpappas:
> Are you suggesting not installing Ansible with pip install ansible-core but using something like zypper in ansible ?
Yes and not just for ansible. 
In general I believe our first instict should be to use everything that comes from our repos and then use external sources for the rest.

mpagot:
> > Are you suggesting not installing Ansible with pip install ansible-core but using something like zypper in ansible ?
> 
> Yes and not just for ansible. In general I believe our first instinct should be to use everything that comes from our repos and then use external sources for the rest.

I think that this change can potentially facilitate doing `zypper in ansible-core` in the future:  `ansible-core` is smaller and simpler that the big `ansible` package.

PR #190, opened 2023-11-16
Task no longer installing particular version of gcc

a-kpappas:
In sap-hana-preconfigure.yaml the "Ensure minimum version of GCC10 are installed on SLE12" is improperly trying to install a particular version of gcc instead of defining a minimum version like the SLE15 task.

This removes the check for OS version as gcc>=10 is enough for all OSes

Verification Runs:
12SP5 & 15SP4 http://openqaworker15.qa.suse.cz/tests/overview?build=fake_build


mpagot (approved):
LGTM

PR #192, opened 2023-11-21
Add Ansible yamllint

mpagot:
Fix Makefile about all the Ansible targets, they now catch all relevant files.
Fix all the warning reported by yamllint.
Add a github action to lint the Ansible code with yamllint.

Ticket : TEAM-7184

Verification: is the pipeline execution in this PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,11 @@ static-flake8:
 	@find scripts -type f -not -path "scripts/qesap/.tox/*" -not -path "scripts/qesap/.venv/*" -name \*.py -exec flake8 --ignore=E501 {} +
 
 static-ansible-yaml:
-	@find ansible -type f -iname "*.yaml" -or -iname "*.yaml" -exec yamllint {} +
+	@tools/ansible_yaml_lint
 
+static-ansible-syntax: export ANSIBLE_ROLES_PATH=tools/dummy_roles
 static-ansible-syntax:
-	@find ansible/playbooks -type f -iname "*.yaml" -maxdepth 1  -exec ansible-playbook -l all -vvvv -i terraform/azure/inventory.yaml --syntax-check  {} +
+	@find ansible/playbooks/ -type f -iname "*.yaml" -maxdepth 1  -exec ansible-playbook -l all -i tools/inventory.yaml --syntax-check  {} +
 
 static-ansible-lint:
 	@ansible-lint ansible/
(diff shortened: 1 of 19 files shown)


Makefile
```diff
@@ -23,10 +23,10 @@ static-flake8:
 	@find scripts -type f -not -path "scripts/qesap/.tox/*" -not -path "scripts/qesap/.venv/*" -name \*.py -exec flake8 --ignore=E501 {} +
 
 static-ansible-yaml:
-	@find ansible -type f -iname "*.yaml" -or -iname "*.yaml" -exec yamllint {} +
+	@tools/ansible_yaml_lint
 
 static-ansible-syntax:
-	@find ansible/playbooks -type f -iname "*.yaml" -maxdepth 1  -exec ansible-playbook -l all -vvvv -i terraform/azure/inventory.yaml --syntax-check  {} +
+	@find ansible/playbooks/ -type f -iname "*.yaml" -and ! -name "registration_role.yaml" -maxdepth 1  -exec ansible-playbook -l all -vvvv -i terraform/azure/inventory.yaml --syntax-check  {} +
```
BillAnastasiadis, line 29: In the future, if we add more roles, we should find a better solution for this

mpagot: yes, some more context:
- `static-ansible-syntax` is not executed as github action at each PR yet. 
- `static-ansible-syntax` produce many warnings at the moment that prevent it to be immediately added as github action, otherwise all PR will be blocked
- the specific reason that I decided to ignore `registration_role.yaml` is that ansible-syntax check fails on it as it needs a role from another repo to be "installed". We need to find a way to manage this situation (how to lint a file that needs roles from a different repo). From another point of view (changing lint with  a human user of this repo) it is like to say that user has to read https://github.com/SUSE/qe-sap-deployment#using-roles-from-a-different-repo before to use this playbook.

BillAnastasiadis (approved):
LGTM

Makefile
BillAnastasiadis: Agree

alvarocarvajald, line 29: The other check (the one in `static-ansible-yaml`) calls a script with only one `find` (and some variables). Wouldn't it make sense to also turn this check into a script in `tools/`?

alvarocarvajald (approved):
LGTM. Just a question.

Makefile
mpagot: I decided to focus this PR on yamllint and leave all the work about `ansible-playbook ``syntax-check` for a next PR. Yes I agree it make sense to move this one too to a dedicated script.

PR #193, opened 2023-11-24
Add ansible playbook syntax check

mpagot:
Call ansible-playbook for each playbook using --syntax-check. 
Dummy roles and inventory is provided to avoid false error detection. 
The python script also format the error output to be compliant to github pipeline standard.
Add Github pipeline to run this test.

Ticket : TEAM-7184

Verification: is the pipeline execution in this PR

Here how an error will looks like
image

### DIFF
--- a/.github/workflows/ansible.yml
+++ b/.github/workflows/ansible.yml
@@ -12,7 +12,12 @@ jobs:
       - name: Install all requirements
         run: |
           python3 -m pip install --upgrade -r requirements.txt
+          ansible-galaxy install -r requirements.yml
 
       - name: Run ansible-yaml-lint
         run: |
            make static-ansible-yaml
+
+      - name: Run ansible-playbooks syntax-check
+        run: |
+           make static-ansible-syntax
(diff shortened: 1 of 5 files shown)


alvarocarvajald (approved):
LGTM

.github/workflows/ansible.yml
```diff
@@ -12,7 +12,12 @@ jobs:
       - name: Install all requirements
         run: |
           python3 -m pip install --upgrade -r requirements.txt
+          ansible-galaxy install -r requirements.yml
```
mpagot, line 15: What is interesting is that calling ansible-playbook also need al Ansible dependency to be ok. So this test is also an indirect test for the Ansible dependency.

.gitignore
```diff
@@ -3,7 +3,7 @@ venv/
 **/.venv/
 **/__pycache__/
 secret/
-inventory.yaml
+terraform/**/inventory.yaml
```
mpagot, line 6: Needed not to ignore the dummy inventory in the tools folder

PR #208, opened 2024-02-07
Validate Terraform 1.5.7 and add tool validation pipeline

mpagot:
# Commit description

Improve the Dockerfile to use Terraform 1.5.7 and Python 3.11.
Add virtual env in the Dockerfile to fix a bug that prevented pip installation of python packages.
Add bash script to validate the docker image.
Add githubaction to build the container image and run the validation
test.

# Context

These project needs some tools to be pre-installed on the JumpHost to perform the deployment. Two of them are Python and Terraform. There's no way for the project to have control over what is installed, so tools version for these two are only suggested in the project main readme.

Dockerfile is not the main way this deployment is used, openQA is not running the `qe-sap-deployment` code from within a container.

# PR purpose

This PR is about:
 - creating a test bench to validate external dependency not directly under control of this project (Terraform and Python)
 - test created in this PR are not enough to say that a new version of Python or Terraform are good to be used with `qe-sap-deployment` ...
 - ... but it is a good starting point :1st_place_medal: 

# Pro n Cons
- What I do not like about this PR is that version of Terraform and Python are spread in many places:
    - In the main readme
    - Dockerfile
    - In some YAML to define the GH pipeline

If something fails during the execution of the new pipeline is not so easy to understand what was wrong

The new pipeline could fails due to external issues, the Dockerfile are installing stuff from the web

# Ticket
 TEAM-9050

# Verification
VR is the PR pipeline, took to the one named `tools` https://github.com/SUSE/qe-sap-deployment/actions/runs/7822845007/job/21342714076?pr=208

### DIFF
--- a/.github/workflows/ansible.yml
+++ b/.github/workflows/ansible.yml
@@ -1,4 +1,5 @@
 name: Ansible
+
 on: [push, pull_request]
 
 jobs:
--- a/.github/workflows/gluescript.yml
+++ b/.github/workflows/gluescript.yml
@@ -2,6 +2,8 @@ name: Glue Script
 
 on:
   push:
+    branches:
+      - "main"
     paths:
       - scripts/qesap/**
       - .github/workflows/gluescript.yml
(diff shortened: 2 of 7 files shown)


BillAnastasiadis (approved):
LGTM

lpalovsky (approved):
LGTM

PR #222, opened 2024-04-26
WIP-EXAMPLE: Add json output

BillAnastasiadis:
DO NOT MERGE this is just an example showcasing things that need to be done to use json output for ansible in qe-sap deployment.

IMPORTANT NOTES: 
- The .cfg file changes are not relevant, as the cfg is not used, but rather the output is configured through env variables
- The cfg is not used anywhere, I believe. This can change if we want to use the `ansible.cfg` for stuff, by adding 
```
    original_env['ANSIBLE_CONFIG'] = os.path.join(base_project, 'ansible.cfg')
```
in cmds.py.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,2 +1,6 @@
 [ssh_connection]
 pipelining = True
+[defaults]
+stdout_callback = json
+callback_whitelist = json
+json_indent = 4
--- a/scripts/qesap/lib/cmds.py
+++ b/scripts/qesap/lib/cmds.py
@@ -289,8 +289,12 @@ def ansible_command_sequence(configure_data_ansible, base_project, sequence, ver
     ansible_cmd_seq.append({'cmd': ssh_share})
     original_env = dict(os.environ)
     original_env['ANSIBLE_PIPELINING'] = 'True'
-    if profile:
-        original_env['ANSIBLE_CALLBACK_WHITELIST'] = 'ansible.posix.profile_tasks'
+    original_env['ANSIBLE_CALLBACKS_ENABLED'] = 'json'
+    original_env['ANSIBLE_LOAD_CALLBACK_PLUGINS'] = '1'
+    original_env['ANSIBLE_STDOUT_CALLBACK'] = 'json'
+    #original_env['ANSIBLE_CONFIG'] = os.path.join(base_project, 'ansible.cfg')
+    #if profile:
+        #original_env['ANSIBLE_CALLBACK_WHITELIST'] = 'json'
     if 'roles_path' in configure_data_ansible:
         original_env['ANSIBLE_ROLES_PATH'] = configure_data_ansible['roles_path']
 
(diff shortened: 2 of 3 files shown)


ansible.cfg
```diff
@@ -1,2 +1,6 @@
 [ssh_connection]
-pipelining = True
\ No newline at end of file
+pipelining = True
+[defaults]
+stdout_callback = json
+callback_whitelist = json
+json_indent = 4
```
BillAnastasiadis, line 6: These changes do not matter, as the cfg file is not used. Look at the description to use it in the future

scripts/qesap/lib/cmds.py
```diff
@@ -289,8 +289,12 @@ def ansible_command_sequence(configure_data_ansible, base_project, sequence, ver
     ansible_cmd_seq.append({'cmd': ssh_share})
     original_env = dict(os.environ)
     original_env['ANSIBLE_PIPELINING'] = 'True'
-    if profile:
-        original_env['ANSIBLE_CALLBACK_WHITELIST'] = 'ansible.posix.profile_tasks'
+    original_env['ANSIBLE_CALLBACKS_ENABLED'] = 'json'
+    original_env['ANSIBLE_LOAD_CALLBACK_PLUGINS'] = '1'
+    original_env['ANSIBLE_STDOUT_CALLBACK'] = 'json'
```
BillAnastasiadis, line 294: These are the ones to use in order to have json output

scripts/qesap/lib/cmds.py
```diff
@@ -345,18 +349,15 @@ def cmd_ansible(configure_data, base_project, dryrun, verbose, destroy=False, pr
         return Status("ok")
 
     inventory = os.path.join(base_project, 'terraform', configure_data['provider'], 'inventory.yaml')
-    ansible_cmd_seq = ansible_command_sequence(configure_data['ansible'], base_project, sequence, verbose, inventory, profile)
+    ansible_cmd_seq = ansible_command_sequence(configure_data['ansible'], base_project, sequence, False, inventory, profile)
```
BillAnastasiadis, line 352: Removed the verbose output, not sure if it impacts json output in any way, I think not

scripts/qesap/lib/process_manager.py
```diff
@@ -31,6 +32,11 @@ def subprocess_run(cmd, env=None):
...
     stdout = [line.decode("utf-8") for line in proc.stdout.splitlines()]
 
+    print(f"OUTPUT START FOR COMMAND: {cmd}")
+    sys.stdout.flush()
     for line in stdout:
-        log.debug('OUTPUT: %s', line)
+        print(line)
+        sys.stdout.flush()
+    print(f"OUTPUT END FOR COMMAND: {cmd}")
+    sys.stdout.flush()
```
BillAnastasiadis, line 41: This part needed to change, otherwise the logging messages were mixed with the json output. This was strange and shouldn't happen, since this whole execution is serial and not parallel, but I think that the `print` commands stacked up instead of being immediately delivered to stdout, so at some point some of the latter logging messages were mixed in. 

This was fixed by using sys.stdout.flush() after each line printing, to send it to stdout immediatelly, so... I was probably right? Dunno. But it works.

scripts/qesap/lib/cmds.py
hadeskun: Maybe not but will afect this pr already merge https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/18787

scripts/qesap/lib/process_manager.py
hadeskun: nice master!!

PR #225, opened 2024-05-06
Update qe-sap-deployment internal testing

mpagot:
More selective gh pipeline for PR
Document how to run pytest in VSCode
Improve the Makefile: exclude generated files, add e2e testsrun py target through tox when possible, update KICS, add Terraform linting. Remove the old legacy build.sh.
Keep only one requirements-dev and update tools in it. Add ansible binary detectio in the gluescript (fix UT about it too)


mpagot:
Same code added as part of other PR

PR #230, opened 2024-05-10
Update some of the tools used during code validation

mpagot:
Update dev requirements.txt
Get rid of the old bash deployment scripts
Reorganize the Makefile, add `beyond` target for the experimental stuff


alvarocarvajald (approved):
LGTM

PR #242, opened 2024-06-17
Remove the duplicated requirements-dev.txt

mpagot:
The correct one is newr to the glue script main file


alvarocarvajald (approved):
LGTM

PR #243, opened 2024-06-17
Remove terraform duplicated logs

mpagot:
Remove from the qesap.py stdout some logs that report messages from Terraform twice. Issue is present only when run the glue script in verbose mode.


alvarocarvajald (approved):
LGTM

PR #247, opened 2024-06-24
WIP: Add playbook for openqa pt install

BillAnastasiadis:
**DO NOT MERGE**


ansible/playbooks/ptf_installation.yaml
```diff
@@ -5,21 +5,43 @@
...
+    - name: Include external variables
+      ansible.builtin.include_vars: ./vars/hana_media.yaml
+
     - name: Create directory for PTF installation
       file:
         path: "{{ ptf_dir }}"
         state: directory
         mode: '0755'
 
-    - name: Download PTF files recursively with wget
```
mpagot, line 19: is it possible to find a way not to trough away this part and to let have both to survive in the same playbooks (or two playbooks if needed)

ansible/playbooks/ptf_installation.yaml
```diff
@@ -5,21 +5,43 @@
...
-      command: "wget --no-directories --recursive --reject 'index.html*' --user={{ ptf_user }} --password={{ ptf_password }} --no-parent {{ ptf_url }}"
-      args:
-        chdir: "{{ ptf_dir }}"
+    - name: Convert comma-separated string to list
+      set_fact:
+        az_blobs: "{{ ptf_files | split(',') }}"
+
+    - name: Extract the first part of az_container_name
+      set_fact:
+        az_container_top: "{{ az_container_name.split('/')[0] }}"
```
mpagot, line 29: we can avoid this "trick" (that create hidden dependency of path used in hana_media.yaml and playbooks/sap-hana-download-media.yaml) and just force the user to provide the path in a second variable just for this playbook

```
ptf_installation.yaml -e ptf_container="aaa/bbb/ptf" -e ptf_files="aaa.rpm,bbb.rpm"
```

or with the path in each file element
```
ptf_installation.yaml  -e ptf_files="aaa/bbb/ptf/aaa.rpm,aaa/bbb/ptf/bbb.rpm"
```

ansible/playbooks/ptf_installation.yaml
BillAnastasiadis: I just did it with 2 playbooks. I think it can be done in 1 by checking which vars are available, but it's also fine like this. What do you think?

ansible/playbooks/ptf_installation.yaml
mpagot: We can decide to use a dedicated container for PTF.
Maybe one with a subfolder for each PTF

ansible/playbooks/ptf_installation.yaml
mpagot: considering number of tasks in common I'd prefer 1 configurable playbook:
- If user provide `ptf_user,ptf_password,ptf_url` is manual
- If user provide `ptf_file` is the other

ansible/playbooks/ptf_installation.yaml
```diff
@@ -5,21 +5,48 @@
...
+      delay: "{{ url_retries_delay }}"
+      with_items: "{{ az_blobs }}"
+      when: sas_token is defined
+
     - name: Download PTF files recursively with wget
       command: "wget --no-directories --recursive --reject 'index.html*' --user={{ ptf_user }} --password={{ ptf_password }} --no-parent {{ ptf_url }}"
       args:
         chdir: "{{ ptf_dir }}"
+      when:
+        - sas_token is not defined
```
mpagot, line 46: I think you can remove this condition `- sas_token is not defined` from here. The other 3 are enough. Or is like this to not allow, in any case, that some var combination allow both download method to be executed in the same execution?

ansible/playbooks/ptf_installation.yaml
```diff
@@ -5,21 +5,48 @@
   become: true
   become_user: root
   vars:
-    ptf_url: "your_ptf_url"
-    ptf_user: "your_username_here"
-    ptf_password: "your_pass_here"
+    url_timeout: 30
```
mpagot, line 8: why we like them to be parametric?
why hardcoding value directly in `Download PTF files with SAS token` is not enough?

mpagot (approved):
LGTM

My only concern is that it come out with a complicated API (no idea if /how we can improve it)

url_timeout, url_retries_cnt, url_retries_delay   can be configured from the external but has default. There's no validatio nof values if configured from the external.


sas_token is optional. Its presence is one of the variables checked to decide which download task to use. Name overlap with the one used in hana_media.yaml but they are two different variables (even if for the moment, due to what we currently have in Azure, token is the same). It cold be an idea to rename it to ptf_sas_token and just externally (openQA) use same token used for HANA install?

ptf_files is mandatory if sas_token is defined. It has to have a special format like comma separated list of filename (or is filename + some path?). Its presence is not checked or validated before to start using it

container, storage are mandatory if sas_token is defined. They have to be strings. Their presence are not checked or validated before to start using them

ptf_user, ptf_password, ptf_url are optional and used to select a specific download task. The first two are string, password is a secret, ptf_url is an url. None are validated.

ansible/playbooks/ptf_installation.yaml
BillAnastasiadis: exactly - just in case someone calls the playbook with all possible optional args, then it will only use the token method and not both

BillAnastasiadis:
> LGTM
> 
> My only concern is that it come out with a complicated API (no idea if /how we can improve it)
> 
> * `url_timeout`, `url_retries_cnt`, `url_retries_delay`   can be configured from the external but has default. There's no validatio nof values if configured from the external.
> * `sas_token` is optional. Its presence is one of the variables checked to decide which download task to use. Name overlap with the one used in hana_media.yaml but they are two different variables (even if for the moment, due to what we currently have in Azure, token is the same). It cold be an idea to rename it to `ptf_sas_token` and just externally (openQA) use same token used for HANA install?
> * `ptf_files` is mandatory if `sas_token` is defined. It has to have a special format like comma separated list of filename (or is filename + some path?). Its presence is not checked or validated before to start using it
> * `container`, `storage` are mandatory if `sas_token` is defined. They have to be strings. Their presence are not checked or validated before to start using them
> * `ptf_user`, `ptf_password`, `ptf_url` are optional and used to select a specific download task. The first two are string, password is a secret, ptf_url is an url. None are validated.

- this was inherited from `sap_download-hana-media.yaml`, which uses them in the `Download HANA media with SAS token` task. Since the sas-token method of downloading in my playbook is basically copied from that function, I used the same args and values.
- `sas_token` does not have the same name as in `sap_download-hana-media.yaml`: there, it's `az_sas_token`, while I renamed it to simply `sas_token`
- I can validate it, should I try? Or should we let the task fail and the reviewer review it?
- Similar question as above
- Similar as above

lpalovsky (approved):
LGTM

PR #254, opened 2024-07-05
Add e2e test for junit and profile

mpagot:
Ticket [TEAM-9475](https://jira.suse.com/browse/TEAM-9475)

What I did:
- create a new playbook that has a test using `ansible.builtin.pause`. It add 30sec to the time profile
- create a new config.yaml to use the new playbook
- add a test step to call the gluescript with `--profile` and grep for the profile string in the output
- add a test to call the gluescript with `--junit` and look for .xml junit reports


mpagot:
new test added via other PR

PR #263, opened 2024-08-29
Add az_key_name support at glue script level only

mpagot:
Add validation and hana_media.yaml write of this new parameter. In a second commit it will be used to generate the SAS token in Ansible.

Ticket: TEAM-9540

# Verification

## HanaSR

 - sle-15-SP6-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_6_BYOS-qesap_aws_r48xlarge_test@64bit -> http://openqaworker15.qa.suse.cz/tests/295703

 # qesap

 - sle-15-SP5-HanaSr-Aws-Byos-x86_64-Build15-SP5_2024-08-29T02:03:22Z-hanasr_aws_test_fencing_sbd@64bit -> http://openqaworker15.qa.suse.cz/tests/295704

### DIFF
--- a/scripts/qesap/lib/cmds.py
+++ b/scripts/qesap/lib/cmds.py
@@ -58,6 +58,8 @@ def create_hana_media(config_ansible, apiver):
     hanamedia_content['az_container_name'] = config_ansible['az_container_name']
     if 'az_sas_token' in config_ansible:
         hanamedia_content['az_sas_token'] = config_ansible['az_sas_token']
+    if 'az_key_name' in config_ansible:
+        hanamedia_content['az_key_name'] = config_ansible['az_key_name']
     hanamedia_content['az_blobs'] = config_ansible['hana_media']
     return hanamedia_content, None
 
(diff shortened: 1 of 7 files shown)


scripts/qesap/lib/cmds.py
```diff
@@ -58,6 +58,8 @@ def create_hana_media(config_ansible, apiver):
     hanamedia_content['az_container_name'] = config_ansible['az_container_name']
     if 'az_sas_token' in config_ansible:
         hanamedia_content['az_sas_token'] = config_ansible['az_sas_token']
+    if 'az_key_name' in config_ansible:
+        hanamedia_content['az_key_name'] = config_ansible['az_key_name']
```
BillAnastasiadis, line 62: This could be an `elif`, but maybe it's better this way - conflicts will be handled in the playbook to help with manual deployments too

mpagot: No, I intentionally place `if`. I like the final `hana_media.yaml` to potentially has both, and at least one.

This way I think Ansible cold have all the freedom deciding what to do:
Ansible can (my preferred) if `az_sas_token` is provided, just use it and ignore `az_key_name`. If token is not provided calculate it on the fly been sure that needed argument `az_key_name` is in available as input.

alvarocarvajald (approved):
LGTM

BillAnastasiadis (approved):
LGTM

PR #264, opened 2024-08-30
Move token generation to ansible

BillAnastasiadis:
This pr enables qe-sap-deployment to generate the SAS token in ansible when it is not defined in the config.yaml. It requires providing the key name for the generation of the token. It can be used both for manual deployments (by providing the key name instead of the generated key) and openqa runs (if the key is provided as a setting instead of the generated SAS token).

- Related ticket: https://jira.suse.com/browse/TEAM-9540
- Verification Runs: 
**With token generated in openqa and '`QESAPDEPLOY_HANA_KENAME` not set:**
https://openqaworker15.qa.suse.cz/tests/295828
**Without token generated in openqa and `QESAPDEPLOY_HANA_KEYNAME` set as an openqa variable:**
https://openqaworker15.qa.suse.cz/tests/296013

### DIFF
--- a/ansible/playbooks/sap-hana-download-media.yaml
+++ b/ansible/playbooks/sap-hana-download-media.yaml
@@ -23,6 +23,48 @@
         group: root
         mode: 0755
 
+    - name: Retrieve account key
+      ansible.builtin.command: >-
+        az storage account keys list \
+        --account-name {{ az_storage_account_name }} \
+        --query "[?contains(keyName,'{{ az_key_name }}')].value" \
+        -o tsv
+      delegate_to: 127.0.0.1
+      no_log: true
+      run_once: true
+      register: az_account_key
+      when: az_sas_token is not defined or az_sas_token == ""
+
+    - name: "Set expiry"
+      ansible.builtin.command: "date -u +'%Y-%m-%dT%H:%MZ' -d '+3 hours'"
+      delegate_to: 127.0.0.1
+      run_once: true
+      register: expiry
+      when: az_sas_token is not defined or az_sas_token == ""
+
+    - name: Generate SAS token
+      ansible.builtin.command: >-
+        az storage container generate-sas \
+          --account-name {{ az_storage_account_name }} \
+          --account-key {{ az_account_key.stdout }} \
+          --name {{ az_container_name.split('/')[0] }} \
+          --permission r \
+          --expiry {{ expiry.stdout }} \
+          --out tsv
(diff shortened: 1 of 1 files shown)


alvarocarvajald:
Seems *QESAPDEPLOY_HANA_KENAME* was set in both verification runs. Not sure if that has an impact though.

ansible/playbooks/sap-hana-download-media.yaml
```diff
@@ -23,6 +23,48 @@
...
+        --query "[?contains(keyName,'{{ az_key_name }}')].value" \
+        -o tsv
+      delegate_to: 127.0.0.1
+      no_log: true
+      run_once: true
+      register: az_account_key
+      when: az_sas_token is not defined or az_sas_token == ""
+
+    - name: "Set expiry"
+      ansible.builtin.command: "date -u +'%Y-%m-%dT%H:%MZ' -d '+20 minutes'"
```
alvarocarvajald, line 39: I guess 20 minutes is enough in a normal deployment, but what happens when it fails and we need to re-deploy? A new token will be generated?

alvarocarvajald (approved):
LGTM. But I have a question.

mpagot:
> Seems **QESAPDEPLOY_HANA_KEYNAME** was set in both verification runs. Not sure if that has an impact though.

It's fine to do... after this PR we can move to a new situation where we only have QESAPDEPLOY_HANA_KEYNAME

ansible/playbooks/sap-hana-download-media.yaml
mpagot: Good point! We could make it configurable. On the other hand token is not needed across all the test or even not across all the deployment. Token just needs to be valid during installer download Ansible task in same playbook

mpagot (approved):
LGTM

ansible/playbooks/sap-hana-download-media.yaml
BillAnastasiadis: Hmmm you are correct. I will increase it and run a final VR, merging automatically if it passes since there are 2 approvals

mpagot:
We should also update the documentation and example files to explain and suggesting the usage of the new key approach. One file to be updated is https://github.com/SUSE/qe-sap-deployment/pull/263/files#diff-4cd7c4907a755b045fd47cc82f36e9d7f1b39cd595156a707502729220d5398aR49

PR #261, opened 2024-08-28
Remove machine type default from AWS

mpagot:
Terraform AWS code variables has default about machine type in two place:
- HIGH: in the main variables.tf
- LOW: in each module variable.tf It is problematic to maintain these two level of defaults with the same values.
Removing defaults from the HIGH level result in all machine types variables to become mandatory in the terraform.tfvars and so in the conf.yaml.

This commit remove defaults from LOW level. Doing so does not change user experience as HIGH levels will always provide value to LOW level from user tfvars file or default.
This commit also ensure that each description field is at the first place if available.

Ticket https://jira.suse.com/browse/TEAM-7678

# Verification
- sle-15-SP5-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_5_BYOS-qesap_aws_saptune_test@64bit -> http://openqaworker15.qa.suse.cz/tests/295667
 - sle-15-SP6-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_6_BYOS-qesap_aws_r48xlarge_test@64bit -> http://openqaworker15.qa.suse.cz/tests/295668
 - sle-15-SP6-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_6_BYOS-qesap_aws_r58xlarge_test@64bit -> http://openqaworker15.qa.suse.cz/tests/295669
 - sle-15-SP5-HanaSr-Aws-Byos-x86_64-Build15-SP5_2024-08-29T02:03:22Z-hanasr_aws_test_fencing_sbd@64bit -> http://openqaworker15.qa.suse.cz/tests/295670

PR #265, opened 2024-09-02
Use vm_size in AWS modules

mpagot:
Try to use a uniform naming convention for modules across different cloud providers.
This PR intentionally only change internal modules API and does not touch `terraform/aws/variables.tf`.
In this way the end user has not to change anything in the conf.yaml.
Remove some vm_size default at module level, only to have defaults in the upper layer and in a single place.
Reorder variables fields always to have description at first place.

Ticket: https://jira.suse.com/browse/TEAM-7678

# Verification

 ## sle-15-SP6-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_6_BYOS-qesap_aws_r48xlarge_test@64bit
 
- http://openqaworker15.qa.suse.cz/tests/296104 :green_circle:  Variable file http://openqaworker15.qa.suse.cz/tests/296104/file/configure-terraform.tfvars has `hana_instancetype = "r4.8xlarge"` and it result in http://openqaworker15.qa.suse.cz/tests/296104/logfile?filename=deploy-terraform.plan.log.txt to have 

```
# module.hana_node.aws_instance.hana[0] will be created
  + resource "aws_instance" "hana" {
    ...
      + instance_type                        = "r4.8xlarge"
```

##  sle-15-SP6-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_6_BYOS-qesap_aws_r5bmetal_test@64bit 
 - http://openqaworker15.qa.suse.cz/tests/296105 :green_circle: it has `hana_instancetype = "r5b.metal"` in http://openqaworker15.qa.suse.cz/tests/296105/file/configure-terraform.tfvars and the same used in http://openqaworker15.qa.suse.cz/tests/296105/logfile?filename=deploy-terraform.plan.log.txt

### DIFF
--- a/terraform/aws/modules/drbd_node/variables.tf
+++ b/terraform/aws/modules/drbd_node/variables.tf
@@ -18,40 +18,39 @@ variable "drbd_count" {
   default     = 2
 }
 
-variable "instance_type" {
+variable "vm_size" {
   description = "The instance type of drbd node"
   type        = string
-  default     = "t2.large"
 }
 
 variable "availability_zones" {
-  type        = list(string)
   description = "Used availability zones"
+  type        = list(string)
 }
 
 variable "vpc_id" {
-  type        = string
   description = "Id of the vpc used for this deployment"
+  type        = string
 }
 
 variable "subnet_address_range" {
-  type        = list(string)
   description = "List with subnet address ranges in cidr notation to create the netweaver subnets"
+  type        = list(string)
 }
 
 variable "key_name" {
-  type        = string
   description = "AWS key pair name"
+  type        = string
 }
 
 variable "security_group_id" {
-  type        = string
   description = "Security group id"
+  type        = string
 }
 
 variable "route_table_id" {
-  type        = string
   description = "Route table id"
+  type        = string
 }
 
 variable "aws_credentials" {
--- a/terraform/azure/modules/drbd_node/variables.tf
+++ b/terraform/azure/modules/drbd_node/variables.tf
@@ -11,18 +11,27 @@ variable "resource_group_name" {
   type = string
 }
 
+variable "drbd_count" {
+  description = "Number of drbd machines to create the cluster"
+  type        = number
+  default     = 2
+}
+
+variable "vm_size" {
+  description = "The instance type of drbd node"
+  type        = string
+  default     = "Standard_D2s_v3"
+}
+
 variable "network_subnet_id" {
   type = string
 }
 
 variable "storage_account" {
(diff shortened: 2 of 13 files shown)


terraform/aws/modules/drbd_node/variables.tf
```diff
@@ -18,40 +18,39 @@ variable "drbd_count" {
   default     = 2
 }
 
-variable "instance_type" {
+variable "vm_size" {
   description = "The instance type of drbd node"
   type        = string
-  default     = "t2.large"
```
BillAnastasiadis, line 24: Is this default used anywhere, or can it be safely removed?

mpagot: All the 3 cloud providers has the same layered architecture:
- main layer: it has a main and a variable file. The main in this layer take variables in this layer and use them to configure variables for each module in the next layer.
- modules: each module has its main and its variable file

Now use this model about variables to configure the VM type:
- **AZURE**: 
  1. Example of top level variable https://github.com/SUSE/qe-sap-deployment/blob/f3e2b04a9af57e5eba2004fde18f13949b2e33b3/terraform/azure/variables.tf#L202 . It has a default.
  2. Example of top level main that configure module layer low level https://github.com/SUSE/qe-sap-deployment/blob/f3e2b04a9af57e5eba2004fde18f13949b2e33b3/terraform/azure/main.tf#L204
  3. Low level variable: https://github.com/SUSE/qe-sap-deployment/blob/f3e2b04a9af57e5eba2004fde18f13949b2e33b3/terraform/azure/modules/hana_node/variables.tf#L62  It still has a default, but we can think about removing it (like I started doing in AWS https://github.com/SUSE/qe-sap-deployment/pull/261/files#diff-5e0f222194e1a8dbb339d9203a8a956ba75c2d44405c68ebd4435807f01a369eL23)
  4. Low level main where the variable value is used when finally create the VM https://github.com/SUSE/qe-sap-deployment/blob/f3e2b04a9af57e5eba2004fde18f13949b2e33b3/terraform/azure/modules/hana_node/main.tf#L392 .

terraform/azure/modules/drbd_node/variables.tf
```diff
@@ -11,18 +11,27 @@ variable "resource_group_name" {
   type = string
 }
 
+variable "drbd_count" {
```
mpagot, line 14: Just reorder variables making this file more similar to the AWS one

PR #266, opened 2024-09-03
Run aws set_facts conditionally

BillAnastasiadis:
`Set corosync facts` in `cluster_bootstrap.yaml` is being run for both aws and gcp, but only applies to aws, uses an aws-specific variable and that causes gcp tests to fail due to missing variable. This pr conditionally runs the task, only if platform is aws.

- Related ticket: https://jira.suse.com/browse/TEAM-9643
- Verification runs:
http://openqaworker15.qa.suse.cz/tests/296118 (passed the point of trouble, fails due to native fencing not working)
http://openqaworker15.qa.suse.cz/tests/296122 (PASS, sbd implementation)

### DIFF
--- a/ansible/playbooks/tasks/cluster-bootstrap.yaml
+++ b/ansible/playbooks/tasks/cluster-bootstrap.yaml
@@ -222,7 +222,7 @@
     migration_threshold: "{{ (crm_rsc_options_show.stdout | regex_search('migration-threshold=([0-9]*)', '\\1'))[0] | default('false') }}"
     op_default_timeout: "{{ (crm_op_options_show.stdout | regex_search('timeout=([0-9]*)', '\\1'))[0] | default('false') }}"
 
-- name: Set corosync facts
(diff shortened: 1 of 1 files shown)


ansible/playbooks/tasks/cluster-bootstrap.yaml
```diff
@@ -222,7 +222,7 @@
     migration_threshold: "{{ (crm_rsc_options_show.stdout | regex_search('migration-threshold=([0-9]*)', '\\1'))[0] | default('false') }}"
     op_default_timeout: "{{ (crm_op_options_show.stdout | regex_search('timeout=([0-9]*)', '\\1'))[0] | default('false') }}"
 
-- name: Set corosync facts
+- name: Set corosync facts [AWS]
```
mpagot, line 225:
```suggestion
- name: NAtive fencing command [AWS]
```

mpagot (approved):
LGTM

PR #272, opened 2024-09-09
Use smaller disks by default for tests on Azure

zhaocongjia:
Uses smaller disks for hanasr tests.  Tests do not break.  And test run time remains similar.

https://jira.suse.com/browse/TEAM-9625

VRs:

15SP5 BYOS spn:
https://openqa.suse.de/tests/15326917
https://openqa.suse.de/tests/15326918
https://openqa.suse.de/tests/15326920
https://openqa.suse.de/tests/15327230

15SP6 BYOS:
https://openqa.suse.de/tests/15375352

15SP5 BYOS msi:
https://openqa.suse.de/tests/15375353

15SP4 PAYG:
https://openqa.suse.de/tests/15375354

12SP5 PAYG msi:
https://openqa.suse.de/tests/15375355

### DIFF
--- a/terraform/azure/variables.tf
+++ b/terraform/azure/variables.tf
@@ -194,7 +194,7 @@ variable "sles4sap_uri" {
 }
 
 # For reference:
-# Standard_M32ls has 32 VCPU, 256GiB RAM, 1000 GiB SSD
+# Standard_E4s_v3 has 4 VCPU, 32GiB RAM, 64 GiB SSD
 # You could find other supported instances in Azure documentation
 variable "hana_vm_size" {
   description = "VM size for the hana machine"
@@ -212,7 +212,7 @@ variable "hana_data_disks_configuration" {
   type = map(any)
   default = {
     disks_type       = "Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS"
-    disks_size       = "128,128,128,128,128,128,128"
+    disks_size       = "32,32,16,16,32,64,32"
     caching          = "None,None,None,None,None,None,None"
     writeaccelerator = "false,false,false,false,false,false,false"
     # The next variables are used during the provisioning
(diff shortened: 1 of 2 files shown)


lilyeyes (approved):
LGTM.

terraform/azure/variables.tf
```diff
@@ -212,7 +212,7 @@ variable "hana_data_disks_configuration" {
   type = map(any)
   default = {
     disks_type       = "Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS"
-    disks_size       = "128,128,128,128,128,128,128"
+    disks_size       = "32,32,16,16,32,64,32"
```
mpagot, line 215: Maybe we can add a comment in the code to say something like

```
This default is aligned to default for hana_vm_size
```

We can also update comment in line 197 to talk about `Standard_E4s_v3` and not `Standard_M32ls`

mpagot: Each value now is using a different number: does it make sense to document why and which disk has which size? Where we should do it? Here in this file or in the readme?

mpagot:
We should update and expand documentation in https://github.com/SUSE/qe-sap-deployment/blob/9f4aa9c4ba11ce96d56a1f5b9dda603c11d08948/terraform/azure/README.md?plain=1#L209 and https://github.com/SUSE/qe-sap-deployment/blob/9f4aa9c4ba11ce96d56a1f5b9dda603c11d08948/terraform/azure/README.md?plain=1#L224

zhaocongjia:
Good idea, I should keep the code and documentation in sync.

zhaocongjia:
Additional qesap regression VRs:

https://openqaworker15.qa.suse.cz/tests/297047
https://openqaworker15.qa.suse.cz/tests/297046

terraform/azure/variables.tf
zhaocongjia: I think those disks sizes are not finalized.  I'll collect some data about cost saving with this commit.  Then there shall be a follow up ticket to optimize the disk sizes further.  Then we can add definitive documentation about disk sizes.

mpagot (approved):
LGTM

PR #297, opened 2024-11-28
Remove bsc1224797 workaround

mpagot:


alvarocarvajald:
LGTM, but let's avoid merging this while this is still not fixed in 15-SP4, 15-SP5 and 12-SP5.

PR #301, opened 2024-12-04
Add GCP peering to terraform

BillAnastasiadis:
This pr **conditionally** adds peering with IBSM to GCP innside terraform, during resource creation. Nothing else will need to be done for the creation or the destruction of the peering (it will be managed exclusively by terraform). If the control variable `` doesn't exist, it defaults to false, and the deployment continues as before without creating the peering or related resources.

To make use of this and create a deployment with IBSM Peering, user (or openqa) needs to provide the following variables in config.yaml:

```
enable_ibsm_peering: 'true'
ibsm_vpc_name: '<gcp mirror vpc name>'
ibsm_subnet_name: '<gcp mirror subnet name>'
ibsm_subnet_region: '<gcp mirror region>'

```

And that's it. By providing those 4 variables, the hana nodes will be created peered with gcp IBSM, and the peering and related resources will be automaticlly destroyed when terraform destroy is called.

- Verification Run:
https://openqaworker15.qa.suse.cz/tests/305879 (this is without any of the new variables set - so no peering or associated resources are created, and test runs like before)

### DIFF
--- a/terraform/gcp/infrastructure.tf
+++ b/terraform/gcp/infrastructure.tf
@@ -82,3 +82,80 @@ resource "google_compute_firewall" "ha_firewall_allow_tcp" {
   }
 }
 
+# IBSM related network imports
+
+data "google_compute_network" "ibsm_vpc" {
+  count = var.enable_ibsm_peering
+  name  = var.ibsm_vpc_name
+}
+
+data "google_compute_subnetwork" "ibsm_subnet" {
+  count  = var.enable_ibsm_peering
+  name   = var.ibsm_subnet_name
+  region = var.ibsm_subnet_region
+}
+
+# IBSM Peering resources
+
+# Peering from HANA VPC to ibsm VPC
+resource "google_compute_network_peering" "hana_to_ibsm" {
+  count = var.enable_ibsm_peering
+
+  name                 = "${local.deployment_name}-hana-to-ibsm"
+  network              = local.network_link
+  peer_network         = data.google_compute_network.ibsm_vpc[0].self_link
+  export_custom_routes = true
+  import_custom_routes = true
+}
+
+# Peering from ibsm VPC to HANA VPC
+resource "google_compute_network_peering" "ibsm_to_hana" {
+  count = var.enable_ibsm_peering
+
+  name                 = "ibsm-to-${local.deployment_name}-hana"
+  network              = data.google_compute_network.ibsm_vpc[0].self_link
+  peer_network         = local.network_link
+  export_custom_routes = true
+  import_custom_routes = true
+}
+
+# Allow internal traffic from HANA VPC to ibsm VPC
+resource "google_compute_firewall" "allow_internal_from_hana" {
+  count = var.enable_ibsm_peering
+
+  name    = "${local.deployment_name}-fw-allow-from-hana"
+  network = data.google_compute_network.ibsm_vpc[0].name
+
+  allow {
+    protocol = "icmp"
+  }
+
+  allow {
+    protocol = "tcp"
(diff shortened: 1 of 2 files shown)


a-kpappas (dismissed):

terraform/gcp/infrastructure.tf
```diff
@@ -82,3 +82,90 @@ resource "google_compute_firewall" "ha_firewall_allow_tcp" {
...
+  name    = "${local.deployment_name}-fw-allow-from-ibsm"
+  network = local.vpc_name
+
+  allow {
+    protocol = "icmp"
+  }
+
+  allow {
+    protocol = "tcp"
+    ports    = ["0-65535"]
```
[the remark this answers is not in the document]
BillAnastasiadis: I have no idea tbh 😅 just wanted to be consistent with the deployment. TBH these rules refer strictly to traffic between the peered networks, so it shouldn't really be a problem - the pre-existing hana-subnet open ports may be more of a concern. But that's out of scope for this ticket

terraform/gcp/variables.tf
```diff
@@ -775,3 +775,29 @@ variable "iscsi_remote_python" {
   default     = "/usr/bin/python3"
 }
 
+# Peering related variables
+
+variable "enable_ibsm_peering" {
+  description = "Enable VPC peering between HANA VPC and ibsm VPC"
+  type        = bool
```
mpagot, line 782: it can be `int` with default to zero. Doing so we can also simplify all the 
```
count = var.enable_ibsm_peering ? 1 : 0
```

to 
```
count = var.enable_ibsm_peering
```

terraform/gcp/variables.tf
```diff
@@ -775,3 +775,29 @@ variable "iscsi_remote_python" {
...
+variable "ibsm_subnet_name" {
+  description = "Name of the ibsm subnet"
+  type        = string
+  default     = ""
+}
+
+variable "ibsm_subnet_region" {
+  description = "Region of the ibsm subnet"
+  type        = string
+  default     = "europe-west1"
```
mpagot, line 801:
```suggestion
  default     = ""
```

terraform/gcp/infrastructure.tf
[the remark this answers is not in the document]
jankohoutek: If you check the IBSm, only ports needed are 22 (SSH), 80(HTTP) and ICMP (ping) 
There is wide open rule, but only for outbound.
So, it's correct here, but too wide between lines  133-141

jankohoutek (changes requested):

terraform/gcp/infrastructure.tf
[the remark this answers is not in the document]
jankohoutek: Also, there is no need of any UDP traffic, just a TCP and ICMP

terraform/gcp/infrastructure.tf
```diff
@@ -82,3 +82,90 @@ resource "google_compute_firewall" "ha_firewall_allow_tcp" {
...
+  name    = "${local.deployment_name}-fw-allow-from-hana"
+  network = data.google_compute_network.ibsm_vpc[0].name
+
+  allow {
+    protocol = "icmp"
+  }
+
+  allow {
+    protocol = "tcp"
+    ports    = ["0-65535"]
```
jankohoutek, line 135: ports = ["22,80"]

jankohoutek, line 135:
```suggestion
    ports    = ["22,80"]
```

mpagot (approved):
LGTM

PR #302, opened 2024-12-04
Improve YAML to tfvar conversion

mpagot:
Better handling of strings and bool in the terraform section of the conf.yaml


scripts/qesap/test/unit/test_qesap_lib_config.py
```diff
@@ -25,7 +57,7 @@ def test_tfvars_yaml_list(config_data_sample):
     :return:
     true or false
     """
-    hana_ips = ['10.0.0.2', '10.0.0.3']
+    hana_ips = ["10.0.0.2", "10.0.0.3"]
```
lpalovsky, line 60: What is the purpose of double quotes here?

lpalovsky (approved):
LGTM

scripts/qesap/test/unit/test_qesap_lib_config.py
mpagot: At some point I run the file through [Black](https://github.com/psf/black) (formatting tool) that for some reason prefer that. Here an explanation why https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings

lpalovsky: Oh I see. Thanks for explanation.

alvarocarvajald (approved):
LGTM

PR #308, opened 2025-01-20
Replace pacemaker deprecated commands for Azure

BillAnastasiadis:
The warnings about the deprecated `ms` command and the Master/Slave attributes were there from previous versions, but in sp7 they became fatal. This pr is about replacing the deprecated commands with their proposed equivalents.

**- Verification runs:**
15sp7 (should use new way): https://openqa.suse.de/tests/16537074#
_All below should use old way:_
https://openqa.suse.de/tests/16543350#
https://openqa.suse.de/tests/16537127#
https://openqa.suse.de/tests/16537128#
https://openqa.suse.de/tests/16537129#
https://openqa.suse.de/tests/16537131#

PR #310, opened 2025-01-22
Replace depracated crm commands aws/gcp

BillAnastasiadis:
A followup and the aws/gcp equivalent of https://github.com/SUSE/qe-sap-deployment/pull/308

The warnings about the deprecated `ms` command and the Master/Slave attributes were there from previous versions, but in sp7 they became fatal. This pr is about replacing the deprecated commands with their proposed equivalents.

**- Verification runs:** 
http://openqaworker15.qa.suse.cz/tests/311557


mpagot (approved):
LGTM

PR #316, opened 2025-02-07
Use proper default for the iscsi machine type

mpagot:
t3a.small is not a supported size by sles4sap 15sp5. Change to a size that is valid for majority of modern sles4sap.


alvarocarvajald (approved):
LGTM

badboywj:
LGTM, thanks

lilyeyes (approved):
LGTM.

PR #329, opened 2025-03-14
Add more e2e qesap.py tests

mpagot:
## Verification
https://github.com/SUSE/qe-sap-deployment/actions/runs/13857548499/job/38777683133?pr=329#step:10:96

### DIFF
--- a/scripts/qesap/test/e2e/main_local.tf
+++ b/scripts/qesap/test/e2e/main_local.tf
@@ -1,0 +1,4 @@
+resource "local_file" "foo" {
+  content  = "foo!"
+  filename = "${path.module}/foo.bar"
+}
--- a/scripts/qesap/test/e2e/test.sh
+++ b/scripts/qesap/test/e2e/test.sh
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+test_split () {
+   echo "---------------------------------------------------------------------"
+}
+
 test_step () {
   echo "#######################################################################"
   echo "# $1"
@@ -12,7 +16,7 @@ test_die () {
 }
 
 test_file () {
-  [[ -f "$1" ]] || test_die "Generated file $1 not found!"
+  [[ -f "$1" ]] || test_die "Generated file '$1' not found!"
 }
 
 test_step "Initial folder structure cleanup and preparation"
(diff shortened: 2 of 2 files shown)


scripts/qesap/test/e2e/test.sh
```diff
@@ -12,7 +16,7 @@ test_die () {
 }
 
 test_file () {
-  [[ -f "$1" ]] || test_die "Generated file $1 not found!"
+  [[ -f "$1" ]] || test_die "Generated file '$1' not found!"
```
ilmanzo, line 19: does single quotes prevents variable expansion ?

alvarocarvajald: It doesn't, but I don't think Michele was trying to prevent it.

scripts/qesap/test/e2e/test.sh
```diff
@@ -149,18 +170,86 @@ set +e
...
+qesap.py --verbose -b ${QESAPROOT} -c test_3.yaml --dryrun terraform -d || test_die "test_3.yaml fail on dryrun terraform destroy"
+
+test_split
+echo "Run the script again collecting the output"
+qesap.py -b ${QESAPROOT} -c test_3.yaml --dryrun terraform -d |& tee "${THIS_LOG}"
+for t_step in destroy; do
+  echo "***Detect ${t_step} command***"
+  grep -qE "terraform.*${t_step}" \
+    "${THIS_LOG}" || test_die "test_3.yaml terraform dryrun does not have expected ${t_step} command in the output"
+done
```
alvarocarvajald, line 229: Why do we need a loop for a single test step? Is it just thinking ahead?

alvarocarvajald (approved):
LGTM, but I do have 1 question

scripts/qesap/test/e2e/test.sh
mpagot: no, is only to mark that there's something. They are there as in time to time function can be called without argument (with argument that is empty)

Without this change result was
```
"Generated file  not found!"
```

Not it is 
```
"Generated file '' not found!"
```

scripts/qesap/test/e2e/test.sh
mpagot: cut and paste, let me fix it

scripts/qesap/test/e2e/main_local.tf
```diff
@@ -0,0 +1,4 @@
+resource "local_file" "foo" {
```
mpagot, line 1: This is the most cool stuff, it allow to run e2e tests without any cloud credential but simulating resource creation and delete

PR #331, opened 2025-03-19
Update gcloud installer

mpagot:
TEAM-10178 change installer binary usr from google, using one without version specific. Change the way gcloud is tested by giving full binary path: somehow the installation method does not add it to the PATH, maybe as we run installation as root user.


lpalovsky (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #332, opened 2025-03-26
Add azure peering to terraform

BillAnastasiadis:
Following the enablement of direct terraform network peering for GCP IBSM, this ticket adds terraform network peering for Azure.

**How to test:** run a manual deployment for azure and add these variables to config.yaml, in the terraform section:

```
ibsm_vnet_name: <ibsm_vnet_name>
enable_az_vnet_peering: true
```
`ibsm_rg: <ibsm_rg>` is also needed, but already provided in current azure jobs.

Simply run terraform using a yaml with the above added in config.yaml, then ssh to one of the nodes and ping the IBSM private ip.

Related ticket: https://jira.suse.com/browse/TEAM-9885
- Verification runs: 

https://openqa.suse.de/tests/17187386 (using the new way - check the CASEDIR and the new variable `IBSM_VNET`, those changes are enough to trigger the new peering)

https://openqa.suse.de/tests/17187371 (using the old way - no CASEDIR and no `IBSM_VNET` variable provided)

### DIFF
--- a/terraform/azure/infrastructure.tf
+++ b/terraform/azure/infrastructure.tf
@@ -313,3 +313,34 @@ resource "azurerm_network_security_group" "mysecgroup" {
     workspace = local.deployment_name
   }
 }
+
+# IBSM
+data "azurerm_virtual_network" "ibsm_vnet" {
+  count               = (var.ibsm_rg != "" && var.ibsm_vnet_name != "") ? 1 : 0
+  name                = var.ibsm_vnet_name
+  resource_group_name = var.ibsm_rg
+}
+
+# Peering from the IBSM vnet to deployment vnet
+resource "azurerm_virtual_network_peering" "ibsm_to_target" {
+  count = (var.ibsm_rg != "" && var.ibsm_vnet_name != "") ? 1 : 0
+
+  name                      = "${var.ibsm_vnet_name}-${local.vnet_name}"
+  resource_group_name       = var.ibsm_rg
+  virtual_network_name      = data.azurerm_virtual_network.ibsm_vnet[0].name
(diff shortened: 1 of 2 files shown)


terraform/azure/variables.tf
```diff
@@ -881,3 +881,22 @@ variable "iscsi_remote_python" {
   default     = "/usr/bin/python3"
 }
 
+# Network peering variables
+variable "enable_az_vnet_peering" {
```
mpagot, line 885: Is it strictly needed? Cannot we use `len(ibsm_rg) != 0` in place of it? Do we need a variable to expose to the external and not just infere indirectly if user like to create peering from presence of absence of other variables?

terraform/azure/infrastructure.tf
```diff
@@ -313,3 +313,34 @@ resource "azurerm_network_security_group" "mysecgroup" {
...
+  count               = var.enable_az_vnet_peering ? 1 : 0
+  name                = var.ibsm_vnet_name
+  resource_group_name = var.ibsm_rg
+}
+
+# Peering from the IBSM vnet to deployment vnet
+resource "azurerm_virtual_network_peering" "ibsm_to_target" {
+  count = var.enable_az_vnet_peering ? 1 : 0
+
+  name                      = "${var.ibsm_vnet_name}-to-${local.vnet_name}"
```
mpagot, line 328: Let's try this one to match https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/09e4331173c538f956294da7d788098205da7911/lib/qesapdeployment.pm#L2066

```
my $peering_name = "$source_vnet-$target_vnet";
```

mpagot (approved):
LGTM

PR #338, opened 2025-03-31
Fix dryrun output for Ansible

mpagot:
Dryrun is mind to print Terraform and Ansible commands that a user can
directly run. The first Ansible command produced an invalid command due
to missing quotation. This commit is about fixing this issue. The
approach is to change the internal rappresentation for the commands,
from a list of list of string (subprocess expect a list of strings for a
single command), to a list of strings: each command is just rappresented
with a single space separated string. This format can be directly used
for the dryrun mode. When not in dryrun mode, process_manager is using
shlex.split to produce the input subproces needs.

### DIFF
--- a/scripts/qesap/lib/cmds.py
+++ b/scripts/qesap/lib/cmds.py
@@ -173,26 +173,29 @@ def cmd_terraform(configure_data, base_project, dryrun, workspace='default', des
     if not cfg_paths:
         return Status(f"Invalid folder structure at {base_project}")
 
+    terraform_common_cmd = f"{config.get_terraform_bin()} -chdir={cfg_paths['provider']}"
+
     cmds = []
-    terraform_common_cmd = [config.get_terraform_bin(), f"-chdir={cfg_paths['provider']}"]
     if destroy:
-        cmds.append(terraform_common_cmd + ['destroy', '-auto-approve', '-no-color'])
+        cmds.append(f"{terraform_common_cmd} destroy -auto-approve")
         if workspace != 'default':
-            cmds.append(terraform_common_cmd + ['workspace', 'select', 'default', '-no-color'])
-            cmds.append(terraform_common_cmd + ['workspace', 'delete', workspace, '-no-color'])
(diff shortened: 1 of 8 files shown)


scripts/qesap/lib/cmds.py
```diff
@@ -173,26 +173,29 @@ def cmd_terraform(configure_data, base_project, dryrun, workspace='default', des
...
     for command in cmds:
+        command += ' -no-color'
         if dryrun:
-            print(' '.join(command))
+            print(command)
         else:
             ret, out = lib.process_manager.subprocess_run(command)
             log.debug("Terraform process return ret:%d", ret)
-            log_filename = f"terraform.{command[2]}.log.txt"
+            log_filename = f"terraform.{command.split()[2]}.log.txt"
```
alvarocarvajald, line 198: How was this working before? Was `command` a list ... doesn't look like it from L191.

alvarocarvajald (approved):
LGTM

scripts/qesap/lib/cmds.py
mpagot: It was `command = ['terraform', 'init']` and this PR change it to `command = 'terraform init'`

scripts/qesap/lib/process_manager.py
```diff
@@ -11,19 +12,20 @@
...
+                      before to be used as input for subprocess.run
     Returns:
         (int, list of string): exit code and list of stdout
     """
     if 0 == len(cmd):
         log.error("Empty command")
         return (1, [])
 
-    log.info("Run:       '%s'", ' '.join(cmd))
+    log.info("Run:       '%s'", cmd)
```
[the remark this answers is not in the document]
mpagot: Not a big deal here as we just print a simple string variable, but not using f-string in logging is intentional. 

There's some literature about it, maybe the most official one is https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-not-lazy.html

PR #351, opened 2025-04-24
Stop using no more supported Ubuntu vergion in gluescript workflow

mpagot:
Move to ubuntu-latest like in all theother workflows.


alvarocarvajald (approved):
LGTM

PR #357, opened 2025-05-15
Fix dynamic nvme-sdX disk errors in r5

BillAnastasiadis:
There is a problem with the displayed device names in r5 instances. While the volumes are described and named in https://github.com/SUSE/qe-sap-deployment/blob/581256196a2c235300c15a20e85248fa55d03a52/terraform/aws/modules/hana_node/variables.tf#L95, their names are assigned on boot time as `nvmeXn1`, where X is and integer from 0 to 7.

The nvmeXn1 displayed device names in r5 instances are assigned randomly: there is no guarantee that nvme0n1 will be the root, or that nvme1n1 will be the first entry in the `hana_data_disks_configuration` (from the hana node's variables.tf). This becomes a problem, because we explicitly set roles for volumes using their nvme names [in a dedicated file ](https://github.com/SUSE/qe-sap-deployment/blob/main/ansible/playbooks/vars/aws_hana_storage_profile.yaml).

This pr exports the volume serials from terraform and uses those serials to pair the original sdX names (as defined [here](https://github.com/jankohoutek/qe-sap-deployment/blob/74b7c043be894bad3ec5415ba26e2f4c25c3f2af/terraform/aws/modules/hana_node/variables.tf#L95)) to their nvmeXn1 counterparts.

- Related ticket: https://jira.suse.com/browse/TEAM-10362
- Verification Runs:
https://openqaworker15.qa.suse.cz/tests/324508
https://openqaworker15.qa.suse.cz/tests/324417
(they fail due to timeout, unrelated to pr)
r4 instance: https://openqa.suse.de/tests/17710895 (fails diue to peering trouble which will be solved by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/22104)

### DIFF
--- a/ansible/roles/qe_sap_storage/tasks/generic_tasks/prepare_storage.yml
+++ b/ansible/roles/qe_sap_storage/tasks/generic_tasks/prepare_storage.yml
@@ -4,28 +4,65 @@
   ansible.builtin.debug:
     msg: "{{ item }} "
 
-- name: Determine the root device
-  ansible.builtin.command: "findmnt -n -o SOURCE /"
-  register: root_device
-  changed_when: false
-
-# Cut any partition suffixes like p1, p2
-- name: Normalize root device name
-  ansible.builtin.set_fact:
-    root_device_name: "{{ root_device.stdout | regex_replace('p[0-9]+$', '') }}"
-
-# If root is not nvme0n1, and it also exists in pv, switch it with nvme0n1 in pv
-# This part of the code only runs if the root_device is not nvme0n1
-# It searches in the item.pv list for the root_device, and if it is found
-# it gets replaced with nvme0n1, which is free to use.
-# This is necessary because nvme1n1-nvme7n1 are given specific roles in aws_hana_storage_profile.yaml
-# and if one of them is root, it needs to be replaced with nvme0n1 (the original root, now free)
-- name: Adjust PV list for storage profile "{{ item.key }}"
-  ansible.builtin.set_fact:
-    adjusted_pv: "{{ item.value.pv | map('regex_replace', '^' ~ root_device_name ~ '$', '/dev/nvme0n1') | list }}"
+# Gather device info (needed to use ansible_facts.devices)
+- name: Gather blockdevice facts (nvmeXn1 sizes)
+  ansible.builtin.setup:
+    gather_subset:
+      - hardware
   when:
-    - root_device_name != '/dev/nvme0n1'
-    - root_device_name in item.value.pv
+    - cloud_platform_is_aws and not aws_machine_type_is_r4
+
+# Get a list of all nvme device names
+- name: Build list of nvme devices and their sizes
+  ansible.builtin.set_fact:
+    nvme_device_list: >-
+      {{
+        ansible_facts.devices |
+        dict2items |
+        selectattr('key', 'match', '^nvme[0-9]+n1$')|
+        list
+      }}
+  when:
+    - cloud_platform_is_aws and not aws_machine_type_is_r4
+
+- name: Convert ebs_id_to_device_name list to a dictionary
+  ansible.builtin.set_fact:
+    ebs_id_to_device_name_map: >-
+      {{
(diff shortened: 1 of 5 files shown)


terraform/aws/inventory.tmpl
```diff
@@ -6,6 +6,7 @@ all:
     cluster_ip: ${cluster_ip}
     aws_stonith_tag: ${stonith_tag}
     aws_region: ${region}
+    ebs_id_to_device_name: ${jsonencode(ebs_map)}
```
mpagot, line 9: interesting it is in `all: --> vars:` section. Considering it is id of disks for a particular VM (like to say are per VM, each of them are specific info about a specific VM) I'd expect to see them in `all: --> children: --> hana: --> hosts:`

terraform/aws/modules/hana_node/outputs.tf
```diff
@@ -29,3 +29,13 @@ output "subnets_by_az" {
     s.availability_zone => s.id
   }
 }
+
+output "volume_id_to_device_name" {
+  value = merge([
+    for inst in aws_instance.hana :
```
mpagot, line 35: this for loops through each vm

terraform/aws/modules/hana_node/outputs.tf
```diff
@@ -29,3 +29,13 @@ output "subnets_by_az" {
...
+    { for bd in inst.ebs_block_device :
```
mpagot, line 36: this for loops through each devices of a particular VM

terraform/aws/modules/hana_node/outputs.tf
```diff
@@ -29,3 +29,13 @@ output "subnets_by_az" {
...
+      bd.volume_id => bd.device_name
```
mpagot, line 37: the `[ ]` is a list of "key value" pairs

terraform/aws/modules/hana_node/outputs.tf
```diff
@@ -29,3 +29,13 @@ output "subnets_by_az" {
...
+    }
+  ]...)
```
mpagot, line 39: what `merge` and the 3 `...` does?

terraform/aws/inventory.tmpl
```diff
@@ -6,6 +6,7 @@ all:
...
   children:
```
mpagot, line 10: here an example of how this line looks like when resolved in an Ansible inventory.yaml file (from PR VRs) https://openqaworker15.qa.suse.cz/tests/324314/logfile?filename=serial_terminal.txt#line-1726

alvarocarvajald (approved):
After discussing this with @BillAnastasiadis over slack, I think I understand enough to say that it LGTM.

terraform/aws/modules/hana_node/outputs.tf
BillAnastasiadis: 3 dots is unpacking the list because the code before creates a list like [key-value1, key-value2 etc], and `merge` gives an error if you try to merge a single list (`like merge([key-value1, key-value2])`)

merge is used to create a single dict out of all the key-value pairs, meaning turn it into a single 
{
key1: value1,
key2: value2
...
}

BillAnastasiadis: This is me trying to create a valid json as final output

ansible/roles/qe_sap_storage/tasks/generic_tasks/prepare_storage.yml
```diff
@@ -4,28 +4,55 @@
   ansible.builtin.debug:
     msg: "{{ item }} "
 
-- name: Determine the root device
-  ansible.builtin.command: "findmnt -n -o SOURCE /"
-  register: root_device
-  changed_when: false
+# Gather device info if needed
```
mpagot, line 7:
```suggestion
# Gather device info  needed for ansible_facts.devices
```

mpagot (approved):
LGTM

PR #363, opened 2025-06-06
GCE: remove unnecessary disk "hana-software"

zhaocongjia:
The `hana-software` disk which mounts to directory `/hana/install`, this disk is unnecessary because it is only used temporarily during installation of HANA, and this disk is not used in other cloud configurations.

https://jira.suse.com/browse/TEAM-10265
VR: https://openqaworker15.qa.suse.cz/tests/328372


badboywj:
LGTM, thanks

PR #367, opened 2025-06-11
Add initial testing files

BillAnastasiadis:
This pr adds the necessary playbooks, roles and support files in qe-sap-deployment to enable testing functionality (specifically the openQA Modules `hana_sr_takeover, `hana_sr_test_secondary` and `qesap_prevalidate`).

Ticket: TEAM-10427

### DIFF
--- a/ansible/playbooks/roles/hana_actions/tasks/main.yml
+++ b/ansible/playbooks/roles/hana_actions/tasks/main.yml
@@ -1,0 +1,179 @@
+---
+- name: Prepare HANA Action - Dump cluster status (crm status)
+  ansible.builtin.command: crm status
+  become: true
+  register: crm_initial
+  changed_when: false
+
+- name: Prepare HANA Action - Assert this node is currently MASTER
+  ansible.builtin.assert:
+    that:
+      - crm_initial.stdout is search('\* Masters:\s*\[\s*{{ inventory_hostname }}\s*\]')
+    fail_msg: "{{ inventory_hostname }} is NOT the master node – aborting test"
+    success_msg: "{{ inventory_hostname }} is the master node"
+
+- name: Prepare HANA Action - Wait cluster idle
+  ansible.builtin.command: cs_wait_for_idle --sleep 5
+  register: idle
+  retries: "{{ cs_wait_timeout // 5 }}"
+  delay: 5
+  become: true
+  changed_when: false
+  until: idle.rc == 0
+
+# SBD related stuff
+- name: Prepare HANA Action - Configure SBD start-delay
+  become: true
+  ansible.builtin.lineinfile:
+    path: /etc/sysconfig/sbd
+    regexp: '^SBD_DELAY_START='
+    line: "SBD_DELAY_START={{ sbd_delay_start | default('yes') }}"
+  when: action == 'crash' or (action == 'stop' and cloud_platform_name == 'EC2')
+
+- name: Prepare HANA Action - Gather SBD-delay parameters   # noqa: command-instead-of-shell the command variable actually reuires shell
+  become: true
+  ansible.builtin.shell: "{{ item.cmd }}"
+  args:
+    warn: false
+  register: sbddelay
+  changed_when: false
+  failed_when: false
+  loop:
+    - name: corosync_token
+      cmd: "corosync-cmapctl | awk -F ' = ' '/totem.token/ {print int($2/1000)}'"
+    - name: corosync_consensus
+      cmd: "corosync-cmapctl | awk -F ' = ' '/totem.consensus/ {print int($2/1000)}'"
+    - name: pcmk_delay_max
+      cmd: "crm resource param stonith-sbd show pcmk_delay_max | sed 's/[^0-9]*//g'"
+    - name: sbd_watchdog_timeout
+      cmd: "grep -oP '(?<=^SBD_WATCHDOG_TIMEOUT=)[[:digit:]]+' /etc/sysconfig/sbd"
+    - name: sbd_delay_start
+      cmd: "grep -oP '(?<=^SBD_DELAY_START=)([[:digit:]]+|yes|no)+' /etc/sysconfig/sbd"
+  when: action == 'crash' or (action == 'stop' and cloud_platform_name == 'EC2')
+
+- name: Prepare HANA Action - Set SBD delay facts
+  ansible.builtin.set_fact:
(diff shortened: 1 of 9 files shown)


ansible/roles/hana_actions/tasks/main.yml
```diff
@@ -0,0 +1,164 @@
...
+  async: 1
+  poll: 0
+  when: action == 'crash'
+
+# Post hana action
+- name: Post HANA Action - Wait SSH back (stop/crash)
+  wait_for_connection:
+    delay: 15
+    timeout: 900
+  when: action in ['crash','stop']
```
mpagot, line 135: why only crash and stop?

ansible/roles/hana_actions/tasks/main.yml
```diff
@@ -0,0 +1,164 @@
...
+- name: Prepare HANA Action - Wait cluster idle
+  command: cs_wait_for_idle --sleep 5
+  register: idle
+  retries: "{{ cs_wait_timeout // 5 }}"
+  delay: 5
+  become: true
+  until: idle.rc == 0
+
+# SBD related stuff
+- name: Prepare HANA Action - Configure SBD start-delay
```
mpagot, line 24: SBD delay tasks:
- it  is so complex and involving for the SUT
- all of the tasks are under the relatively complex condition `when: action == 'crash' or (action == 'stop' and cloud_platform_name == 'EC2')`


I propose to move it under a dedicated role, all the tasks in th erole are without the `when` condition, we only have the condition once when we decide if to call the role

ansible/roles/hana_checks/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+- name: HANA check - Wait cluster idle again
+  become: true
+  command: cs_wait_for_idle --sleep 5
+  register: idle2
+  retries: "{{ cs_wait_timeout // 5 }}"
+  delay: 5
+  until: idle2.rc == 0
+
+- name: HANA check - Gather topology JSON
```
mpagot, line 18: here you say JSON ...

ansible/roles/hana_checks/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  become: true
+  command: SAPHanaSR-showAttr --format=script
```
mpagot, line 20: ... but here you are using `script` format.

I know not all the versions of `SAPHanaSR-showAttr` are supporting json format but it will be interesting to have a task using it, maybe side by side to the one using `script`

ansible/roles/hana_secondary_actions/tasks/main.yml
```diff
@@ -0,0 +1,183 @@
...
+- name: Prepare secondary Action - Wait cluster idle (pre-action)
+  command: cs_wait_for_idle --sleep 5
+  become: true
+  register: idle_pre
+  retries: "{{ cs_wait_timeout // 5 }}"
+  delay: 5
+  until: idle_pre.rc == 0
+
+# SBD-related stuff
+- name: Prepare secondary Action - Gather SBD-delay parameters
```
mpagot, line 24: is it not mostly the same code of https://github.com/SUSE/qe-sap-deployment/pull/367/files#diff-b434aa0a9d87f197afcd15b33e320b9c9f51001932a6cb13ad4a0f9c776bb506R32 ?

If it is so, suggestion in https://github.com/SUSE/qe-sap-deployment/pull/367/files#r2140352482 become even more relevant.

ansible/roles/hana_secondary_actions/tasks/main.yml
```diff
@@ -0,0 +1,183 @@
...
+      cmd: "corosync-cmapctl | awk -F ' = ' '/totem.token/ {print int($2/1000)}'"
+    - name: corosync_consensus
+      cmd: "corosync-cmapctl | awk -F ' = ' '/totem.consensus/ {print int($2/1000)}'"
+    - name: pcmk_delay_max
+      cmd: "crm resource param stonith-sbd show pcmk_delay_max | sed 's/[^0-9]*//g'"
+    - name: sbd_watchdog_timeout
+      cmd: "grep -oP '(?<=^SBD_WATCHDOG_TIMEOUT=)[[:digit:]]+' /etc/sysconfig/sbd"
+    - name: sbd_delay_start
+      cmd: "grep -oP '(?<=^SBD_DELAY_START=)([[:digit:]]+|yes|no)+' /etc/sysconfig/sbd"
+  when: action == 'crash'
```
mpagot, line 43: why this one does not have condition on EC2 ?

ansible/roles/hana_checks/tasks/main.yml
BillAnastasiadis: ah, yes - this didn't work and I switched to script. Let me fix

ansible/roles/hana_checks/tasks/main.yml
BillAnastasiadis: It's not portable, but we can pass `use_angi` to ansible as a variable. I already have a part of the code in `secondary` that uses it optionally (if it exists):

```
master_resource_type: "{{ use_angi | default(false) | bool | ternary('mst','msl') }}"
```

ansible/roles/hana_secondary_actions/tasks/main.yml
BillAnastasiadis: >> Are we poring an OSADO bug? :-)

Yes... :p in this initial port, I change nothing (or at least try to - there are a few changes here and there as a result of me trying to dig into called functions and dependencies)

ansible/playbooks/roles/hana_actions/tasks/main.yml
```diff
@@ -0,0 +1,165 @@
...
+# SBD related stuff
+- name: Prepare HANA Action - Configure SBD start-delay
+  become: true
+  ansible.builtin.lineinfile:
+    path: /etc/sysconfig/sbd
+    regexp: '^SBD_DELAY_START='
+    line: "SBD_DELAY_START={{ sbd_delay_start | default('yes') }}"
+  when: action == 'crash' or (action == 'stop' and cloud_platform_name == 'EC2')
+
+- name: Prepare HANA Action - Gather SBD-delay parameters   # noqa: command-instead-of-shell the command variable actually reuires shell
```
mpagot, line 33:
```suggestion
- name: Prepare HANA Action - Gather SBD-delay parameters   # noqa: command-instead-of-shell the command variable actually requires shell
```

lpalovsky (approved):
LGTM

PR #372, opened 2025-06-24
Add ansible-lint to PR validation

mpagot:


mpagot:
Was here only to demonstrate the ansible-lint workflow behavior

PR #373, opened 2025-06-24
Add ansible-lint test to PR workflows

mpagot:
Add ansible-lint execution only on ansible files modified by the PR. It does not means it will show and block the merge only about warning introduced by the PR itself, but also about all existing warnings in the file that the PR is editing.

Ticket: https://jira.suse.com/browse/TEAM-7184

Example of PR that is changing an Ansible file https://github.com/SUSE/qe-sap-deployment/pull/372


BillAnastasiadis (approved):
LGTM

PR #374, opened 2025-06-30
Tune ansible lint configuration

mpagot:
Let's run the github workflow to use a version of ansible-lint that is compatible to the official ansible-core version used by the project. Adjust yamllint configuration to what used by ansible. Lower fatal ansible-lint rules about roles path: it is something about the qe-sap-deployment structure that needs more works to be addressed.

Ticket: https://jira.suse.com/browse/TEAM-7184

### DIFF
--- a/.github/workflows/ansible.yml
+++ b/.github/workflows/ansible.yml
@@ -21,7 +21,8 @@ jobs:
       - name: Install all requirements
         run: |
           python3 -m pip install --upgrade -r requirements.txt
-          python3 -m pip install ansible-lint==25.6.1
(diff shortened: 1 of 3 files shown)


.github/workflows/ansible.yml
```diff
@@ -21,7 +21,8 @@ jobs:
       - name: Install all requirements
         run: |
           python3 -m pip install --upgrade -r requirements.txt
-          python3 -m pip install ansible-lint==25.6.1
+          # 25.1.3 is the newes version that support ansible-core 2.16.8 we are using right now
```
lpalovsky, line 24: s/newes/newest/

lpalovsky (approved):
LGTM

PR #452, opened 2026-03-31
Remove unused package ha-cluster-bootstrap from deployment

a-kpappas:
While working on TEAM-11037, we noticed that we are not using the scripts in the ha-cluster-bootstrap package on any CSP and any SLE version, so the deployment should not have it as a dependency.

This removes it and makes use of the builtin package management task for ansible instead of the community one.

Related Ticket: TEAM-11105
Verification Runs: https://openqaworker15.qe.prg2.suse.org/tests/overview?build=apappas&distri=sle

### DIFF
--- a/ansible/playbooks/tasks/aws-cluster-bootstrap.yaml
+++ b/ansible/playbooks/tasks/aws-cluster-bootstrap.yaml
@@ -26,18 +26,9 @@
     - credentials
     - config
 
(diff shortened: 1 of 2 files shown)


mpagot (approved):
LGTM

badboywj:
LGTM, thanks

PR #453, opened 2026-04-01
Use system python in the TW container

a-kpappas:
The github pipeline for the qe-sap-deployment project fails at the creation of the Tumbleweed container wtih the message Package 'python312-pip' not found.

This happen because we explicitly chose to python3.12 instead of using the system python3, that automatically changes as the container gets regenerated. This fixes that by not specifying a python version.

Related Ticket: TEAM-11115


mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #461, opened 2026-04-16
Stop pinning azure-core

a-kpappas:
For SLES 12, we needed to force a downgrade of python-azure-core according to https://www.suse.com/support/kb/doc/?id=000020716. But a fix has since been released and we are still using the old package.

This removes the pinning.

Related Ticket: TEAM-11103
VR: https://openqaworker15.qe.prg2.suse.org/tests/363111


mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #462, opened 2026-04-17
Remove explicit installation of python-httplib2

a-kpappas:
TEAM-9693 added an ansible task to explicitly install the gce_fence dependency python-httplib2, that should have been pulled in automatically but wasn't due to bsc#1231153.

The bug has since been resolved and the new task should no longer be needed so we remove it with this commit.

Related Ticket: TEAM-11094
Verification Runs: 
GCP: https://openqaworker15.qe.prg2.suse.org/tests/363266/logfile?filename=Cleanup_resources-hana0-rpm-qa.txt&filter=python-httplib2
AWS: https://openqaworker15.qe.prg2.suse.org/tests/363110/logfile?filename=Crash_site_a-primary-hana0-rpm-qa.txt&filter=python-httplib2 
It is not there in AWS but is not necessary, so we did need to install it.

### DIFF
--- a/ansible/playbooks/tasks/aws-cluster-bootstrap.yaml
+++ b/ansible/playbooks/tasks/aws-cluster-bootstrap.yaml
@@ -35,16 +35,6 @@
   retries: 3
   delay: 60
 
-# This is a workaround for https://bugzilla.suse.com/show_bug.cgi?id=1231153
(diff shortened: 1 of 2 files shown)


mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #469, opened 2026-05-07
Add sleeps between disk attachments

BillAnastasiadis:
This PR adds timeouts between the creation/attachment of disk attachments to the VM in azure, since parallelism sporadically fails which leads to `ConflictingConcurrentWriteNotAllowed` when attachments try to attache to the disk.

(extra info: https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep)

Related ticket: https://jira.suse.com/browse/TEAM-11065

- Verification run: https://openqa.suse.de/tests/22224376

### DIFF
--- a/terraform/azure/modules/hana_node/main.tf
+++ b/terraform/azure/modules/hana_node/main.tf
@@ -20,6 +20,7 @@ locals {
 
   hana_lb_rules_ports_secondary = local.create_active_active_infra == 1 ? local.hana_lb_rules_ports : toset([])
   hostname                      = var.common_variables["deployment_name_in_hostname"] ? format("%s-%s", var.common_variables["deployment_name"], var.name) : var.name
+  disk_attachment_delay         = coalesce(var.disk_attachment_delay, "10s")
 }
 
 resource "azurerm_availability_set" "hana-availability-set" {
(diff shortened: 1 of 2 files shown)


alvarocarvajald (approved):
LGTM

mpagot (approved):
LGTM

terraform/azure/modules/hana_node/main.tf
```diff
@@ -20,6 +20,7 @@ locals {
 
   hana_lb_rules_ports_secondary = local.create_active_active_infra == 1 ? local.hana_lb_rules_ports : toset([])
   hostname                      = var.common_variables["deployment_name_in_hostname"] ? format("%s-%s", var.common_variables["deployment_name"], var.name) : var.name
+  disk_attachment_delay         = coalesce(var.disk_attachment_delay, "10s")
```
mpagot, line 23: Is it `coalesce` needed?

BillAnastasiadis: Just in case our code returns something like disk_attachment_delay="" in tfvars. I remember some similar cases in the past. Theoretically, if we know that we'll never do such a thing, this could be omitted.

PR #473, opened 2026-06-09
Harden hana_prevalidate selectattr filters against missing keys

### LINKED ISSUE #472: hana_prevalidate: selectattr filters crash on sparse SAPHanaSR-angi JSON output
## Summary

The `hana_prevalidate` role aborts with a Jinja `'dict object' has no attribute 'lss'` error whenever `SAPHanaSR-showAttr --format=json` (the SAPHanaSR-angi path) returns a `Site` entry that does not contain every attribute the selectors assume to be present. The same fragility affects the `srPoll`-based selectors a few lines later.

The result is a hard failure at template-evaluation time, before any assertion runs — so the role cannot turn a degraded cluster state into a meaningful assertion failure; it instead fails with a stack trace that does not point at the real condition.

## Affected code

`ansible/playbooks/roles/hana_prevalidate/tasks/main.yaml`

- L144-147 — `Get legacy site lss status`
- L149-152 — `Get modern site lss status`
- L154-157 — `Get site srPoll status`

All three call `selectattr(<key>, …)` over `hana_prevalidate_hana_sr_attrs.Site.values()` without first checking that `<key>` is defined on every value.

## Reference run

- openQA job: https://openqaworker15.qe.prg2.suse.org/tests/372088/
- Failing Ansible log: https://openqaworker15.qe.prg2.suse.org/tests/372088/file/test_cluster-qesap_exec_ansible_test.log.txt

Excerpt of the failure (`Get modern site lss status`, role line 149):

```
fatal: [vmhana01]: FAILED! => {"msg": "The task includes an option with an undefined variable.
The error was: 'dict object' has no attribute 'lss'. 'dict object' has no attribute 'lss' ..."}
```

## Root cause

`SAPHanaSR-showAttr --format=json` (angi) **omits keys whose value is empty**, while `--format=script` (legacy) emits them as blank columns. The legacy-script branch additionally normalizes via `filter_plugins/saphana_parser.py` (`create_final_topology_from_script`), which synthesizes `lss` for every site from `node_state`. The JSON branch performs no equivalent normalization, so any site whose `lss` (or `srPoll`, `srr`, …) is empty appears with the key entirely absent.

When `selectattr('lss', …)` then iterates `Site.values()`, the missing attribute raises immediately — the filter has no opportunity to inspect the value.

## Why this is a robustness problem, independent of any specific cluster state

Whenever angi's JSON happens to omit one of the attributes the selectors read — for any reason, transient or persistent — the role:

1. Produces an uninformative Jinja traceback pointing at the `set_fact` task rather than the actual cluster condition.
2. Skips the downstream assertions (`Assert all HANA sites are online`, `Assert exactly one PRIM site exists`, `Assert correct number of SOK sites exist`), which were designed exactly to surface these states with descriptive messages.
3. Cannot be diagnosed from the failure message alone; it requires inspecting the raw `hana_prevalidate_hana_sr_attrs` to understand what happened.

The role should treat a missing attribute as "this site does not satisfy the predicate" and let the assertions report the real picture.

## Proposed fix

Guard each `selectattr` with `selectattr(<key>, 'defined')` so missing keys are filtered out instead of crashing the template:

```yaml
- name: Get legacy site lss status
  ansible.builtin.set_fact:
    hana_prevalidate_online_sites: >-
      {{ hana_prevalidate_hana_sr_attrs.Site.values()
         | selectattr('lss', 'defined')
         | selectattr('lss', 'in', ['4', 'online']) | list }}
  when: not hana_prevalidate_pacemaker_ge_2_1_7

- name: Get modern site lss status
  ansible.builtin.set_fact:
    hana_prevalidate_online_sites: >-
      {{ hana_prevalidate_hana_sr_attrs.Site.values()
         | selectattr('lss', 'defined')
         | selectattr('lss', 'regex', '^[0-9]+$') | list }}
  when: hana_prevalidate_pacemaker_ge_2_1_7

- name: Get site srPoll status
  ansible.builtin.set_fact:
    hana_prevalidate_prim_sites: >-
      {{ hana_prevalidate_hana_sr_attrs.Site.values()
         | selectattr('srPoll', 'defined') | selectattr('srPoll', 'equalto', 'PRIM') | list }}
    hana_prevalidate_sok_sites: >-
      {{ hana_prevalidate_hana_sr_attrs.Site.values()
         | selectattr('srPoll', 'defined') | selectattr('srPoll', 'equalto', 'SOK') | list }}
```

With this change, an under-populated `Site` entry no longer aborts the play and the existing assertions surface the real state (e.g. `Online sites: 1/2`).

## Optional follow-up

The script-format branch normalizes attributes via the `create_final_topology_from_script` filter. A symmetrical normalization step on the angi-JSON branch (defaulting `lss`/`srPoll`/`srr` to an explicit non-online sentinel when missing) would let the rest of the role stay schema-agnostic and would remove the need for `'defined'` guards in every selector. Out of scope for this issue, but worth considering.

mpagot:
This is a robustness change, not a bug fix: `SAPHanaSR-showAttr --format=json` (SAPHanaSR-angi) omits keys whose value is empty, while the legacy `--format=script` output keeps them as blank columns. As a consequence, a perfectly valid angi response may produce a `Site` entry without `lss`, `srPoll`, etc. The existing `selectattr('lss', ...)` and `selectattr('srPoll', ...)` calls dereference the attribute before testing it.
Guard each selector by chaining `selectattr(<key>, 'defined')` ahead of the real predicate. The Jinja `'defined'` test is the only common test designed to inspect a value without using it.

# Verification
 - sle-15-SP6-Qesap-Aws-Byos-x86_64-BuildLATEST_AWS_SLE15_6-qesap_aws_angi_test -> http://openqaworker15.qe.prg2.suse.org/tests/372221

### DIFF
--- a/ansible/playbooks/roles/hana_prevalidate/tasks/main.yaml
+++ b/ansible/playbooks/roles/hana_prevalidate/tasks/main.yaml
@@ -143,18 +143,18 @@
 
     - name: Get legacy site lss status
       ansible.builtin.set_fact:
-        hana_prevalidate_online_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('lss', 'in', ['4', 'online']) | list }}"
+        hana_prevalidate_online_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('lss', 'defined') | selectattr('lss', 'in', ['4', 'online']) | list }}"
       when: not hana_prevalidate_pacemaker_ge_2_1_7
 
     - name: Get modern site lss status
       ansible.builtin.set_fact:
-        hana_prevalidate_online_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('lss', 'regex', '^[0-9]+$') | list }}"
+        hana_prevalidate_online_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('lss', 'defined') | selectattr('lss', 'regex', '^[0-9]+$') | list }}"
       when: hana_prevalidate_pacemaker_ge_2_1_7
 
     - name: Get site srPoll status
       ansible.builtin.set_fact:
-        hana_prevalidate_prim_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('srPoll', 'equalto', 'PRIM') | list }}"
-        hana_prevalidate_sok_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('srPoll', 'equalto', 'SOK') | list }}"
+        hana_prevalidate_prim_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('srPoll', 'defined') | selectattr('srPoll', 'equalto', 'PRIM') | list }}"
+        hana_prevalidate_sok_sites: "{{ hana_prevalidate_hana_sr_attrs.Site.values() | selectattr('srPoll', 'defined') | selectattr('srPoll', 'equalto', 'SOK') | list }}"
 
     - name: "Assert all HANA sites are online"
       ansible.builtin.assert:


alvarocarvajald (approved):
LGTM

BillAnastasiadis (approved):
LGTM

lpalovsky (approved):
LGTM

PR #475, opened 2026-06-10
Add explicit purge-kernels before reboot

BillAnastasiadis:
This PR adds a task in Ansile to purge the kernel before reboot, instead of waiting for the automatic purging after reboot. This way we can know how much time the purge takes. 

- Related ticket: https://jira.suse.com/browse/TEAM-11124
- Verification Run: https://openqa.suse.de/tests/22774494, https://openqa.suse.de/tests/22775514

### DIFF
--- a/ansible/playbooks/fully-patch-system.yaml
+++ b/ansible/playbooks/fully-patch-system.yaml
@@ -36,6 +36,20 @@
       failed_when: zypper_patch.rc not in [0, 102, 103]
       changed_when: zypper_patch.rc == 102 or zypper_patch.changed
 
+    - name: Proactively run zypper purge-kernels if reboot is requested
+      ansible.builtin.command: zypper purge-kernels
+      register: purge_kernels_result
+      when: zypper_patch.rc in [102, 103]
+      changed_when: true
+      failed_when: false
+
+    - name: Display the time taken for zypper purge-kernels
+      ansible.builtin.debug:
+        msg: "zypper purge-kernels execution time: {{ purge_kernels_result.delta }}"
+      when:
+        - zypper_patch.rc in [102, 103]
+        - purge_kernels_result.rc == 0
+
     - name: Trigger reboot+zypp_wait handler if zypper exit code requires it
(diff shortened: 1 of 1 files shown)


ansible/playbooks/fully-patch-system.yaml
```diff
@@ -36,6 +36,20 @@
       failed_when: zypper_patch.rc not in [0, 102, 103]
       changed_when: zypper_patch.rc == 102 or zypper_patch.changed
 
+    - name: Proactively run zypper purge-kernels if reboot is requested
+      ansible.builtin.command: zypper purge-kernels
+      register: purge_kernels_result
+      when: zypper_patch.rc in [102, 103]
+      changed_when: true
+      failed_when: false
```
BillAnastasiadis, line 44: `failed_when: false` because I don't want to introduce a new point of failure, we still have the task after reboots that waits for the service to finish if something goes wrong.

`changed_when: true` to satisfy the linter - it won't be true only in cases where a reboot is required but without any new kernel, which should be rare (and doesn't really matter).

mpagot: `changed_when: true` is more or less fine I think

mpagot (approved):
LGTM

ansible/playbooks/fully-patch-system.yaml
```diff
@@ -36,6 +36,20 @@
       failed_when: zypper_patch.rc not in [0, 102, 103]
       changed_when: zypper_patch.rc == 102 or zypper_patch.changed
 
+    - name: Proactively run zypper purge-kernels if reboot is requested
+      ansible.builtin.command: zypper purge-kernels
+      register: purge_kernels_result
+      when: zypper_patch.rc in [102, 103]
```
mpagot, line 42: Optionally evaluate if we can find more specific condition than `when: zypper_patch.rc in [102, 103]`

alvarocarvajald (approved):
LGTM

